PR #25, opened 2021-01-21
Ja.w 8767284/uppercase methods

monkey-jeff:
This fixes a bug where our servers would reject methods if the verb wasn't in all caps.  

I am not sure how to write a specific test for this as nock doesn't care about verb case in their matcher.


dbarbato-sf (approved):

kalagarraj:
> This fixes a bug where our servers would reject methods if the verb wasn't in all caps.
> 
> I am not sure how to write a specific test for this as nock doesn't care about verb case in their matcher.

Approving, i think nock has options to pass in a custom matcher..

kalagarraj (approved):

scripts/templateHelpers.ts
```diff
@@ -126,3 +126,12 @@ export const isCommonPathParameter = (property: string): boolean => (property
...
   : false);
+
+/**
+ * Forces the input to all caps.
+ *
+ * @param input - The string you want to change to all caps
+ *
+ * @returns string in all caps
+ */
+export const loud = (input: string): string => String(input).toUpperCase();
```
shirishmaheshwari, line 137: Any reason for calling the variable 'loud'?

shirishmaheshwari (approved):

scripts/templateHelpers.ts
shirishmaheshwari: I see. all caps -> loud :)

bredmond-sf (approved):

monkey-jeff:
> > This fixes a bug where our servers would reject methods if the verb wasn't in all caps.
> > I am not sure how to write a specific test for this as nock doesn't care about verb case in their matcher.
> 
> Approving, i think nock has options to pass in a custom matcher..

I tried that...but the method always got to the matcher in all caps so I couldn't make a negative test for it.

PR #30, opened 2021-02-19
Add option to transform request body

wjhsf:
Added `transformRequest` option to client config. The function accepts `(data, headers)` and can return any valid request body. The default value is a function that converts plain objects to JSON and sets `Content-Type`, but does not modify the body or headers for any other type. This design is a simplified version of the same option in [axios](https://github.com/axios/axios/blob/b7e954eba3911874575ed241ec2ec38ff8af21bb/lib/defaults.js#L31-L55).

Also:
* Changed `ClientConfig` so that instances of the class have a more rigid structure than the initialization options.
* Got rid of `unknown` types for known types, because that's silly.

### DIFF
--- a/src/static/templateUrl.ts
+++ b/src/static/templateUrl.ts
@@ -4,6 +4,8 @@
  * SPDX-License-Identifier: BSD-3-Clause
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
+import type { PathParameters, QueryParameters } from './commonParameters';
+
 export default class TemplateURL extends URL {
   /**
    * @param url -
@@ -13,8 +15,8 @@ export default class TemplateURL extends URL {
     url: string,
     base?: string,
     parameters?: {
-      pathParams?: { [key: string]: unknown },
-      queryParams?: { [key: string]: unknown },
+      pathParams?: PathParameters,
+      queryParams?: QueryParameters,
       origin?: string
     },
   ) {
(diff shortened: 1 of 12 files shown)


bredmond-sf (approved):

monkey-jeff (approved):

src/static/templateUrl.ts
```diff
@@ -49,16 +51,16 @@ export default class TemplateURL extends URL {
...
+  addQueryParams(queryParams?: QueryParameters): void {
     if (queryParams) {
       Object.keys(queryParams).forEach((key) => {
-        if (Array.isArray(queryParams[key])) {
-          const paramArr = queryParams[key] as unknown[];
-          for (let i = 0; i < paramArr.length; i += 1) {
-            this.searchParams.append(key, String(paramArr[i]));
+        const param = queryParams[key];
+        if (Array.isArray(param)) {
+          for (let i = 0; i < param.length; i += 1) {
```
echessman, line 59: I have been told it is more performant to move the .length to a variable outside the for loop definition.

src/test/crossFetchNode.test.ts
```diff
@@ -165,14 +158,12 @@ test('should not fail when arbitrary parameters are configured in fetchOptions',
...
-    ...{
-      fetchOptions: {
-        somekey: 'some value',
-        timeout: 1000,
-      },
-    },
+    fetchOptions: {
+      somekey: 'some value',
+      timeout: 1000,
+    } as any,
```
echessman, line 166: why as any?

gkhunger-sf (approved):

src/test/crossFetchNode.test.ts
wjhsf: Our type definition doesn't allow arbitrary parameters, but that's exactly what this is testing. Using the type assertion `as any` is a workaround.

On a broader note, all we do is pass the options through to `fetch`, unmodified. I don't know what the rationale is for why this test exists. It seems to only be testing the third party dependency, rather than our own code.

PR #34, opened 2021-03-01
Fix shortCode not being set.

wjhsf:
#30 included a refactor that accidentally causes shortCode to not be properly set. This fixes that.

### DIFF
--- a/src/test/crossFetchNode.test.ts
+++ b/src/test/crossFetchNode.test.ts
@@ -61,6 +61,49 @@ test('test getting a token without a proxy', async () => {
   expect(token).toEqual('Bearer test-auth');
 });
 
+test('test getting a token with an invalid short code', async () => {
+  nock('https://invalid-short-code.api.commercecloud.salesforce.com')
+    .post(`/customer/shopper-customers/v1/organizations/${config.parameters.organizationId}/customers/actions/login`)
+    .query({ siteId: config.parameters.siteId, clientId: config.parameters.clientId })
+    .replyWithError('ENOTFOUND-TEST');
+
+  const proxylessConfig: ClientConfigInit = {
+    ...config,
+    parameters: {
+      ...config.parameters,
+      shortCode: 'invalid-short-code',
+    },
+  };
+  delete proxylessConfig.proxy;
+  const client = new ShopperCustomers(proxylessConfig);
+
+  await expect(
+    client.authorizeCustomer({ body: { type: 'guest' } }, true),
+  ).rejects.toEqual({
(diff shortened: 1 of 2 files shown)


src/test/crossFetchNode.test.ts
```diff
@@ -108,7 +108,7 @@ test('should use timeout from fetch options and throw timeout error', async () =
   await expect(client.productSearch({
     parameters: { q: 'sony' }, headers: { authorization: 'Bearer test-auth' },
   })).rejects.toEqual({
-    message: 'network timeout at: https://localhost:3000/search/shopper-search/v1/organizations/ORGANIZATION_ID/product-search?q=sony&siteId=SITE_ID',
+    message: 'network timeout at: https://localhost:3000/search/shopper-search/v1/organizations/ORGANIZATION_ID/product-search?siteId=SITE_ID&q=sony',
```
gkhunger-sf, line 111: why does this matter?

wjhsf: Tests expect URL strings to be an exact match, but query params are not added in a deterministic order. I'm about to file a story to fix this, because it's inconvenient having to constantly tweak the tests.

templates/operations.ts.hbs
```diff
@@ -122,19 +122,24 @@
...
+        version: 'v1',
+        ...configParams
+      } as PathParameters;
+      // Merge in known path parameters from options
       {{#each ../parameters}}
-        "{{{name}}}": optionParams["{{{name}}}"]
-        {{! if it is a common path parameter, fall back to client config }}
-        {{#if (isCommonPathParameter name)}}  ?? configParams["{{{name}}}"]{{/if}}
-        {{#unless @last}},{{/unless}}
+      if (optionParams["{{{name}}}"] !== undefined) {
```
monkey-jeff, line 132: As this is a bug fix where is the accompanying test proving that it fixes the issue?

src/test/crossFetchNode.test.ts
```diff
@@ -40,6 +40,27 @@ test('test getting a token with a post operation', async () => {
   expect(token).toEqual('Bearer test-auth');
 });
 
+test('test getting a token without a proxy', async () => {
```
monkey-jeff, line 43: This is great but would also recommend a negative test when the shortcode isn't set properly and an override test showing you can override the shortcode properly as well.

wjhsf: Added more tests. Now we have coverage for valid short code, invalid short code, and missing short code.

src/test/crossFetchNode.test.ts
```diff
@@ -40,6 +40,70 @@ test('test getting a token with a post operation', async () => {
...
```
monkey-jeff, line 43: +1

monkey-jeff (approved):

monkey-jeff (approved):

echessman (approved):

PR #50, opened 2021-08-02
Package ts declarations

bredmond-sf:
The SDK is generated as Typescript, but the build chain hasn't preserved it to include in the package. There's interest in including it in the package. This adds the rollup plugin to do that and tweaks the configuration without breaking things.
To test this
```
yarn run renderTemplates
yarn run build:lib
```
You will now find `*.d.ts` files in the `lib` folder.  If you run `npm pack`, you will get a `tgz` file that you can `npm install` in another project.


babel.config.js
```diff
@@ -4,19 +4,42 @@
...
-// This file is used by jest
-module.exports = {
-  plugins: ['@babel/plugin-proposal-class-properties'],
-  presets: [
-    [
-      '@babel/preset-env',
-      {
-        targets: {
-          node: 'current',
+const pkg = require('./package.json');
```
bredmond-sf, line 7: It was hard to correctly configure the rollup plugins and babel-jest to use different config files and they are all happier using this one with a the check for test env

package.json
```diff
@@ -138,6 +138,7 @@
     "rollup-plugin-postcss": "3.1.3",
     "rollup-plugin-stylelint": "1.0.0",
     "rollup-plugin-terser": "6.1.0",
+    "rollup-plugin-ts": "^1.4.0",
```
bredmond-sf, line 141: We need this plugin so that we produce a single declaration file to match what rollup produces

rollup.config.js
```diff
@@ -68,11 +69,15 @@ const config = outputs.map(({ file, format }) => ({
       inline: false,
       plugins: postcssPlugins,
     }),
+    ts({
+      transpiler: 'babel',
+      tsconfig: (resolvedConfig) => ({ ...resolvedConfig, noEmit: false }),
```
bredmond-sf, line 74: There's a bug in the react-testing script that it gets angry when we change noEmit in the tsconfig file directly

rollup.config.js
```diff
@@ -68,11 +69,15 @@ const config = outputs.map(({ file, format }) => ({
...
+      exclude: 'node_modules/**',
+    }),
     babel({
       extensions,
       babelHelpers: 'bundled',
       exclude: 'node_modules/**',
-      configFile: './babel.config.rollup.js',
```
bredmond-sf, line 75: Use the default `babel.config.js` instead

babel.config.js
wjhsf: Could still do separate files if you do something like
```ts
return isTest ? require('./babel.config.jest.js') : require('./babel.config.rollup.js');
```
This would keep the two completely separate, which might make things easier to untangle in the future.

Another option with similar abstraction but just one file could be
```ts
return isTest ? getJestConfig() : getRollupConfig();
```

At the very least, you should add comments to explain what each config is used for.

rollup.config.js
wjhsf: This should be a comment in the code, too. 😉

bredmond-sf: done!

babel.config.jest.js
```diff
@@ -0,0 +1,21 @@
+/*
+ * Copyright (c) 2020, salesforce.com, inc.
```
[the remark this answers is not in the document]
bredmond-sf: up-dated!

unandyala (approved):

gkhunger-sf (approved):

wjhsf (approved):

PR #51, opened 2021-08-04
Ka/add user agent header.w 9113466

kalagarraj:
This PR appends custom user agent to user agent header

- Added version template
- Added version file out of rendered templates
- Updated client template to import version template
- Update operations template to utilize version file constants
- Updated util test to ensure sdkVersion is set

### DIFF
--- a/templates/version.ts.hbs
+++ b/templates/version.ts.hbs
@@ -1,0 +1,2 @@
+export const USER_AGENT_HEADER = "user-agent";
+export const USER_AGENT_VALUE = "commerce-sdk-isomorphic@{{metadata.sdkVersion}};"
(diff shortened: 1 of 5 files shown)


wjhsf (approved):
Add integration tests to verify that requests are sent using the custom header.

echessman (dismissed):

wjhsf (dismissed):

templates/version.ts.hbs
```diff
@@ -0,0 +1,2 @@
+export const USER_AGENT_HEADER = "user-agent";
+export const USER_AGENT_VALUE = "commerce-sdk-isomorphic@{{metadata.sdkVersion}}";
```
wjhsf, line 2: The node SDK uses the user agent `commerce-sdk@version;`, with the `;` included in the user agent so that it ends on a known character. Not sure that it's really necessary, but that's what it does, for comparison.

wjhsf (approved):

templates/version.ts.hbs
kalagarraj: since we use semvar, we know the the ending chars are numbers and also user agents are tokenized using space that eliminates the need for a specific ending char...

echessman (approved):

PR #61, opened 2021-12-09
Attempt to demo issue with following redirects client side

johnboxall:
Hey friends,

In https://github.com/salesforceCommerceCloud/pwa-kit, we've been trying to reduce the # of requests made to render any page to improve performance.

In https://github.com/SalesforceCommerceCloud/pwa-kit/pull/254, I'm trying to skip an extra request made in the context of a SLAS login.

The extra request I'm trying to skip is due `fetch`'s default behaviour of following redirect chains.

I'd like to use the `fetchOptions` of `ClientConfig` with the `ShopperLogin` client to disable following redirects:

```js
// This client shouldn't follow redirects.
const loginClient = new ShopperLogin({
  fetchOptions: { redirect: 'manual' },
  ...config
})
```

I've observed that redirects are not followed in Node.js environments, but are in the browser.

This PR demonstrates the issue.

### DIFF
--- a/src/environment/App/App.jsx
+++ b/src/environment/App/App.jsx
@@ -1,3 +1,4 @@
+/* eslint-disable */
 /*
  * Copyright (c) 2021, salesforce.com, inc.
  * All rights reserved.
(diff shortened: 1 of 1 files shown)


src/environment/App/App.jsx
```diff
@@ -7,16 +8,60 @@
...
+import { ShopperLogin, ShopperSearch } from 'lib';
 import config from '../config';
 import './App.css';
 
-const customerClient = new ShopperCustomers(config);
+// For this client, we want to explicitly handle redirects.
+const loginClient = new ShopperLogin({
+  fetchOptions: { redirect: 'manual' },
+  ...config
+})
```
johnboxall, line 21: 1️⃣  – here we create a client and set our intent for it to not follow redirects automajically.

johnboxall:
After reviewing the spec, it appears redirects should _not_ be accessible via the `fetch` API:

> Redirects (a response whose status or internal response’s (if any) status is a redirect status) are not exposed to APIs. Exposing redirects might leak information not otherwise available through a cross-site scripting attack.

https://fetch.spec.whatwg.org/#atomic-http-redirect-handling

So, my original use case won't work in the context of a browser.

For me, one interesting observation was that when using `import fetch from 'cross-fetch` in the browser, you'll get the `whatwg-fetch`'s XHR based implementation.

This was/is surprising to me, because I was under the impression that this would use the native fetch if available ... but it doesn't! which can make reasoning about the behaviour you're seeing tricky.

wjhsf:
> For me, one interesting observation was that when using import fetch from 'cross-fetch in the browser, you'll get the whatwg-fetch's XHR based implementation.
>
> This was/is surprising to me, because I was under the impression that this would use the native fetch if available ... but it doesn't! which can make reasoning about the behaviour you're seeing tricky.

That's a good point! The package [specifically states](https://www.npmjs.com/package/whatwg-fetch#read-this-first) that it's a polyfill. There must be something in our build process that is causing the polyfill to be used regardless of environment.

johnboxall:
@wjhsf for my own learning ;P

`cross-fetch` uses a "ponyfill" approach when imported directly:

```js
// browser.js
import { fetch, Request }  from "cross-fetch";

// Uses XHR 
fetch('https://httpbin.org/anything')
```

https://github.com/lquixada/cross-fetch/blob/main/README.md#install

Which we do right now:

https://github.com/SalesforceCommerceCloud/commerce-sdk-isomorphic/blob/990ac9177f29759946074f1677a5da24e0b0a696/templates/client.ts.hbs#L1

For a polyfilling alternative we could use:

```js
// browser.js
import "cross-fetch/polyfill";

// Uses `fetch` if available, otherwise falls back to XHR
fetch('https://httpbin.org/anything')
```

If there were gaps (features/function) in the XHR based implementation, I think I'd advocate for trying to find a way to get it to use `fetch` in browser if available.

I guess generally I'm also curious about the browser share % of browsers without `fetch`: https://caniuse.com/fetch

It looks like its mainly IE11 which is forcing us to continue polyfilling. IE11 has an EOL date of 2022 June 15 – so it might make sense to revisit this around then.

PR #63, opened 2021-12-10
Add default handler for application/x-www-form-urlencoded request bodies

wjhsf:
* Sets the `Content-Type` header for all requests that have a body.
* Changes the default request transformer to return different values based on the `Content-Type`, rather than trying to guess based on the body itself.
  * `application/json` -> stringified
  * `application/x-www-form-urlencoded` -> `URLSearchParams` instance
* Moves `helpers.ts` to `helpers/types.ts` as prep for my next PR because I didn't clean up before switching to do this one.

This depends on https://github.com/SalesforceCommerceCloud/raml-toolkit/pull/201, so it will be a draft until the next version of raml-toolkit is released.


bredmond-sf (dismissed):

tarcang (approved):

src/static/clientConfig.test.ts
```diff
@@ -67,41 +67,31 @@ describe('ClientConfig constructor', () => {
...
-      expect(headers).toEqual({});
-    });
-
-    test('converts plain objects to JSON and sets Content-Type header', () => {
-      expect(transform({plain: 'object'}, headers)).toBe('{"plain":"object"}');
-      expect(headers).toEqual({'Content-Type': 'application/json'});
-
-      headers = {};
-      expect(transform(Object.create(null), headers)).toBe('{}');
-      expect(headers).toEqual({'Content-Type': 'application/json'});
```
tarcang, line 99: Aren't these test cases still relevant?

tarcang (approved):

src/static/clientConfig.test.ts
wjhsf: They are not. The implementation has changed and no longer does this.

tarcang (approved):

bredmond-sf (approved):

PR #64, opened 2021-12-13
New client config setting: We can throw on every response other than 2xx and 304 responses if throwOnBadResponse is set in ClientConfig.

tarcang:
Note: With the current changes, the throwing will not happen if rawResponse is true.

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,6 +90,10 @@ const config = {
 }
  ```
 
+### Optional settings that can be set in the config object
+*throwOnBadResponse:* Default value is false. When set to true, the sdk with throw on exception on responses with statuses that is not
+2xx and 304.
+
 ## Testing
 
 Two types of tests occur when running `yarn test`. First, unit tests are executed with jest including an enforced coverage level. This is all tests not contained within the sample app path of `src/environment`. If that passes, tests within the sample app path are executed using the `react-scripts` configuration. These allow for testing of the SDK within a sample React application.
--- a/README.md
+++ b/README.md
@@ -91,7 +91,7 @@ const config = {
  ```
 
 ### Optional settings that can be set in the config object
-*throwOnBadResponse:* Default value is false. When set to true, the sdk with throw on exception on responses with statuses that is not
+*throwOnBadResponse:* Default value is false. When set to true, the sdk will throw an Error on responses with statuses that are not
 2xx and 304.
 
 ## Testing
--- a/src/static/clientConfig.ts
+++ b/src/static/clientConfig.ts
@@ -31,6 +31,7 @@ export interface ClientConfigInit<Params extends BaseUriParameters> {
     data: T,
     headers: {[key: string]: string}
   ) => Required<FetchOptions>['body'];
+  throwOnBadResponse?: boolean;
 }
 
 /**
@@ -53,6 +54,8 @@ export default class ClientConfig<Params extends BaseUriParameters>
     ClientConfigInit<Params>['transformRequest']
   >;
 
+  public throwOnBadResponse?: boolean = false;
+
   constructor(config: ClientConfigInit<Params>) {
     this.headers = {...config.headers};
     this.parameters = {...config.parameters};
@@ -70,6 +73,9 @@ export default class ClientConfig<Params extends BaseUriParameters>
     }
     if (config.proxy) {
       this.proxy = config.proxy;
+    }
+    if (config.throwOnBadResponse !== undefined) {
+      this.throwOnBadResponse = config.throwOnBadResponse;
     }
   }
 
--- a/src/static/clientConfig.ts
+++ b/src/static/clientConfig.ts
@@ -74,9 +74,7 @@ export default class ClientConfig<Params extends BaseUriParameters>
     if (config.proxy) {
       this.proxy = config.proxy;
     }
-    if (config.throwOnBadResponse !== undefined) {
-      this.throwOnBadResponse = config.throwOnBadResponse;
-    }
+this.throwOnBadResponse = !!config.throwOnBadResponse;
   }
 
   static readonly defaults: Pick<
--- a/src/static/clientConfig.ts
+++ b/src/static/clientConfig.ts
@@ -11,14 +11,12 @@ import {BaseUriParameters} from './helpers';
  * Alias for `RequestInit` from TypeScript's DOM lib, to more clearly differentiate
  * it from the `RequestInit` provided by node-fetch.
  */
-// eslint isn't aware that we have browser types available, not sure why...
-// eslint-disable-next-line no-undef
 type BrowserRequestInit = RequestInit;
 /**
  * Any properties supported in either the browser or node are accepted.
  * Using the right properties in the right context is left to the user.
  */
-type FetchOptions = BrowserRequestInit & NodeRequestInit;
+export type FetchOptions = NodeRequestInit & BrowserRequestInit;
 
 /**
  * Base options that can be passed to the `ClientConfig` class.
@@ -29,12 +27,8 @@ export interface ClientConfigInit<Params extends BaseUriParameters> {
   headers?: {[key: string]: string};
   parameters: Params;
   fetchOptions?: FetchOptions;
-  // eslint thinks that the names used in the function signature are variables
-  // instead of part of the type, not sure why...
-  transformRequest?: (
-    // eslint-disable-next-line no-unused-vars
-    data: any,
-    // eslint-disable-next-line no-unused-vars
+  transformRequest?: <T>(
+    data: T,
     headers: {[key: string]: string}
   ) => Required<FetchOptions>['body'];
 }
@@ -62,6 +56,7 @@ export default class ClientConfig<Params extends BaseUriParameters>
   constructor(config: ClientConfigInit<Params>) {
     this.headers = {...config.headers};
     this.parameters = {...config.parameters};
+    // shortCode is required in the type, but we still check that it is present for the JS users
     if (!this.parameters.shortCode) {
       throw new Error('Missing required parameter: shortCode');
     }
@@ -88,17 +83,22 @@ export default class ClientConfig<Params extends BaseUriParameters>
      * @param data - Data to transform
      * @returns A JSON string or the unmodified data
      */
-    transformRequest<T>(data: T, headers: {[key: string]: string}): T | string {
+    transformRequest<T>(
+      data: T,
+      headers: {[key: string]: string}
+    ): Required<FetchOptions>['body'] {
+      // NOTE: These type assertions are NOT safe, but I plan on immediately replacing this
+      // implementation in a follow-up PR.
       if (data == null || typeof data !== 'object') {
-        return data;
+        return data as unknown as Required<FetchOptions>['body'];
       }
-      const proto = Object.getPrototypeOf(data);
+      const proto: unknown = Object.getPrototypeOf(data);
       if (Array.isArray(data) || proto === Object.prototype || proto === null) {
         // eslint-disable-next-line no-param-reassign
         headers['Content-Type'] = 'application/json';
         return JSON.stringify(data);
       }
-      return data;
+      return data as unknown as Required<FetchOptions>['body'];
     },
   };
 }
--- a/src/static/clientConfig.ts
+++ b/src/static/clientConfig.ts
@@ -54,7 +54,7 @@ export default class ClientConfig<Params extends BaseUriParameters>
     ClientConfigInit<Params>['transformRequest']
   >;
 
-  public throwOnBadResponse?: boolean = false;
+  public throwOnBadResponse: boolean;
 
   constructor(config: ClientConfigInit<Params>) {
     this.headers = {...config.headers};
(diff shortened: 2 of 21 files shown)


src/test/crossFetchNode.test.ts
```diff
@@ -274,3 +274,58 @@ test('should not fail when arbitrary parameters are configured in fetchOptions',
   });
   expect(response).toEqual({});
 });
+
+test('throws on responses other than and 304 errors', async () => {
```
wjhsf, line 278:
```suggestion
test('throws on error responses', async () => {
```

src/test/crossFetchNode.test.ts
```diff
@@ -274,3 +274,58 @@ test('should not fail when arbitrary parameters are configured in fetchOptions',
...
+    .matchHeader('authorization', 'Bearer test-auth')
+    .reply(400, {}, {'content-type': 'application-json charset=UTF-8'});
+
+  const standardConfig: ClientConfigInit<TestConfigParameters> = {...config};
+  const client = new ShopperSearch(standardConfig);
+  const response = async () =>
+    client.productSearch({
+      parameters: {q: 'sony'},
+      headers: {authorization: 'Bearer test-auth'},
+    });
```
wjhsf, line 296:
```suggestion
  const responsePromise = client.productSearch({
      parameters: {q: 'sony'},
      headers: {authorization: 'Bearer test-auth'},
    });
```
The API call is an async function, so even though you added a `throw` statement, it doesn't throw. It returns a promise that gets rejected. You don't need to wrap the call to test it.

src/test/crossFetchNode.test.ts
```diff
@@ -274,3 +274,110 @@ test('should not fail when arbitrary parameters are configured in fetchOptions',
...
+
+  expect(response).rejects.toThrow();
```
wjhsf, line 298:
```suggestion
  await expect(response).rejects.toEqual(something);
```
1. `expect(promise).rejects` and `expect(promise).resolves` return a Promise, so you must `await` it, otherwise the checks are never actually done.
2. `expect(promise).rejects` basically unwraps the rejected error so that you can make assertions about it. You should make some assertions about it.

src/test/crossFetchNode.test.ts
```diff
@@ -274,3 +274,110 @@ test('should not fail when arbitrary parameters are configured in fetchOptions',
...
+    )
+    .matchHeader('authorization', 'Bearer test-auth')
+    .reply(
+      400,
+      {content: 'not empty'},
+      {'content-type': 'application-json charset=UTF-8'}
+    );
+
+  const standardConfig: ClientConfigInit<TestConfigParameters> = {...config};
+  const client = new ShopperSearch(standardConfig);
```
wjhsf, line 291:
```suggestion
  const client = new ShopperSearch({...config});
```
If you're not doing anything with it, you can just pass it directly. 😉

src/test/crossFetchNode.test.ts
```diff
@@ -274,3 +274,110 @@ test('should not fail when arbitrary parameters are configured in fetchOptions',
...
+  const response = await client.productSearch({
+    parameters: {q: 'sony'},
+    headers: {authorization: 'Bearer test-auth'},
+  });
+
+  expect(response).toEqual({content: 'not empty'});
+});
+
+test('do not throw in empty json body', async () => {
```
wjhsf, line 367: Also add a test case for an empty body.

wjhsf (changes requested):

src/test/crossFetchNode.test.ts
```diff
@@ -274,3 +274,105 @@ test('should not fail when arbitrary parameters are configured in fetchOptions',
...
+    .matchHeader('authorization', 'Bearer test-auth')
+    .reply(
+      400,
+      {content: 'not empty'},
+      {'content-type': 'application-json charset=UTF-8'}
+    );
+
+  const client = new ShopperSearch({...config});
+  const responsePromise = () =>
+    client.productSearch({
```
wjhsf, line 292:
```suggestion
  const responsePromise = client.productSearch({
```
Functions aren't promises. 😉

src/test/crossFetchNode.test.ts
```diff
@@ -274,3 +274,105 @@ test('should not fail when arbitrary parameters are configured in fetchOptions',
...
+      parameters: {q: 'sony'},
+      headers: {authorization: 'Bearer test-auth'},
+    });
+
+  expect(responsePromise).rejects.toEqual(new Error('Error 400: Bad Request'));
```
wjhsf, line 297:
```suggestion
  await expect(responsePromise).rejects.toEqual(new Error('Error 400: Bad Request'));
```
Don't forget to await the expression so that it does what you expect it to!

wjhsf (dismissed):

wjhsf (approved):

templates/operations.ts.hbs
```diff
@@ -180,8 +180,13 @@
       const response = await fetch(req);
       if (rawResponse) {
         return response;
-      } else {
-        return response.json();
+      } else if (response.ok || response.status === 304) {
+        const text = await response.text();
+        // It's ideal to get "{}" for an empty response body, but we won't throw if it's truly empty
+        return text ? JSON.parse(text) : {};
```
wjhsf, line 186: Actually, this may not be what we want to do here. Not all methods necessarily return a value.

wjhsf (changes requested):

templates/operations.ts.hbs
tarcang: Which methods should I not return a value? Is there a doc that describes that?

templates/operations.ts.hbs
```diff
@@ -180,8 +180,15 @@
       const response = await fetch(req);
       if (rawResponse) {
         return response;
-      } else {
-        return response.json();
+      } else if (response.ok || response.status === 304) {
+        {{#unless (eq (getReturnTypeFromOperation this) "void") }}
```
wjhsf, line 184: This if block will be empty when the return type is `void`. I don't think the linter will like that.

tarcang: I guess `return; ` is fine?

src/test/crossFetchNode.test.ts
```diff
@@ -273,3 +273,131 @@ test('should not fail when arbitrary parameters are configured in fetchOptions',
...
+    headers: {
+      authorization: 'Bearer test-auth',
+    },
+    body: {
+      resetToken: 'R1e2s3e4t5T6o7k8e9n0',
+      login: 'janedoe@test.com',
+      newPassword: 'p@assword2',
+    },
+  });
+  expect(response);
```
wjhsf, line 402:
```suggestion
  expect(response).toBeUndefined();
```
(Might not be the actual assertion, but I think it is...)

wjhsf (approved):

src/test/crossFetchNode.test.ts
tarcang: 👌

templates/operations.ts.hbs
```diff
@@ -180,8 +180,17 @@
...
+      } else if (response.ok || response.status === 304) {
+        {{#if (eq (getReturnTypeFromOperation this) "void") }}
+          return;
+        {{else}}
+          const text = await response.text();
+          // It's ideal to get "{}" for an empty response body, but we won't throw if it's truly empty
+          return text ? JSON.parse(text) : {};
+        {{/if}}
+      }
+      else {
```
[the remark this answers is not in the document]
tarcang: 👌

src/test/crossFetchNode.test.ts
tarcang: It is the assertion :)

wjhsf (approved):

templates/operations.ts.hbs
```diff
@@ -115,6 +117,7 @@
         body: {{{getPayloadTypeFromRequest request}}}{{/or}}
       }>,
       rawResponse?: boolean,
+      throwOnBadResponse?: boolean
```
wjhsf, line 120: Throwing on bad responses is probably a behavior that the SDK user wants to be consistent every time.  This should be set on the client config, rather than being passed every single time they call a method.

templates/operations.ts.hbs
```diff
@@ -180,10 +183,26 @@
...
+                throw new Error('Error ' + response.status + ': ' + response.statusText);
+              }
       } else {
-        return response.json();
+        //If rawResponse and throwOnBadResponse is false, go with default behavior.
+        {{#if (eq (getReturnTypeFromOperation this) "void") }}
+          return;
+        {{else}}
+          return response.json();
+        {{/if}}
```
wjhsf, line 204: You can flatten this logic a bit.
```ts
if (rawResponse) {
  return response;
} else if (throwOnBadResponse && !response.ok && response.status !== 304) {
  throw;
} else {
  /* void or JSON block */
}
```

src/test/crossFetchNode.test.ts
```diff
@@ -273,3 +273,201 @@ test('should not fail when arbitrary parameters are configured in fetchOptions',
...
+    );
+
+  const client = new ShopperSearch({...config});
+  const response = await client.productSearch({
+    parameters: {q: 'sony'},
+    headers: {authorization: 'Bearer test-auth'},
+  });
+
+  expect(response).toEqual({content: 'not empty'});
+});
```
wjhsf, line 473: Don't forget to add a test case for `true`!

wjhsf (changes requested):

src/static/clientConfig.ts
```diff
@@ -53,6 +54,8 @@ export default class ClientConfig<Params extends BaseUriParameters>
     ClientConfigInit<Params>['transformRequest']
   >;
 
+  public throwOnBadResponse?: boolean = false;
```
wjhsf, line 57:
```suggestion
  public throwOnBadResponse: boolean;
```
Makes life easier if you have fewer possible states to deal with. 😉 

(Should remain optional in `ClientConfigInit`.)

src/static/clientConfig.ts
```diff
@@ -71,6 +74,9 @@ export default class ClientConfig<Params extends BaseUriParameters>
     if (config.proxy) {
       this.proxy = config.proxy;
     }
+    if (config.throwOnBadResponse !== undefined) {
+      this.throwOnBadResponse = config.throwOnBadResponse;
+    }
```
wjhsf, line 79:
```suggestion
this.throwOnBadResponse = !!config.throwOnBadResponse;
```
Ensures that the property is always a boolean.

templates/operations.ts.hbs
```diff
@@ -180,10 +179,18 @@
...
       } else {
-        return response.json();
+        //If rawResponse and throwOnBadResponse is false, go with default behavior.
+        {{#if (eq (getReturnTypeFromOperation this) "void") }}
+          return;
+        {{else}}
+          const text = await response.text();
+          // It's ideal to get "{}" for an empty response body, but we won't throw if it's truly empty
+          return text ? JSON.parse(text) : {};
+        {{/if}}
```
wjhsf, line 192: After the refactor, this is the last block in the function. The return statement for `void` has become redundant, because there's literally nothing after it. The whole `else` code block can be moved into the non-void return type template block.

src/static/clientConfig.ts
```diff
@@ -31,6 +31,7 @@ export interface ClientConfigInit<Params extends BaseUriParameters> {
     data: T,
     headers: {[key: string]: string}
   ) => Required<FetchOptions>['body'];
+  throwOnBadResponse?: boolean;
```
bredmond-sf, line 34: There is a story to redo the README, but please add a description of this new parameter to to current one so it can easily be moved to the new one!

bredmond-sf (changes requested):

templates/operations.ts.hbs
```diff
@@ -180,10 +179,16 @@
       const response = await fetch(req);
       if (rawResponse) {
         return response;
+      } else if (this.clientConfig.throwOnBadResponse && !response.ok && response.status !== 304) {
+        throw new Error('Error ' + response.status + ': ' + response.statusText);
       } else {
-        return response.json();
+        //If rawResponse and throwOnBadResponse is false, go with default behavior.
```
wjhsf, line 185: What _is_ the default behavior? Is this comment still accurate?

templates/operations.ts.hbs
wjhsf: The entire `else` block should be inside the `{{#unless}}`, otherwise you'll end up with
```ts
else { }
```

README.md
```diff
@@ -90,6 +90,10 @@ const config = {
 }
  ```
 
+### Optional settings that can be set in the config object
+*throwOnBadResponse:* Default value is false. When set to true, the sdk with throw on exception on responses with statuses that is not
```
bredmond-sf, line 94:
```suggestion
*throwOnBadResponse:* Default value is false. When set to true, the sdk will throw an Error on responses with statuses that are not
```

bredmond-sf (dismissed):

wjhsf (dismissed):

joeluong-sfcc (dismissed):

unandyala (dismissed):

wjhsf (approved):

joeluong-sfcc (approved):

PR #66, opened 2021-12-20
W-8856961: Expose the request URIs for API operations

joeluong-sfcc:


src/static/helpers/types.ts
```diff
@@ -43,3 +43,14 @@ export interface QueryParameters {
...
+/* eslint-disable no-shadow */
+// temporary fix for eslint bug: https://github.com/typescript-eslint/tslint-to-eslint-config/issues/856
+/**
+ * Enum type representing different return options
+ */
+export enum returnType {
+  JSON,
+  RESPONSE,
+  REQUEST,
+}
```
wjhsf, line 56:
```suggestion
export type MethodReturnType = "parsed" | "request" | "response";
```
1. So I thought a bit more about this, and I think that a union type is cleaner than an enum here, because that way users can provide just a string, rather than needing to import/use the enum object. [Here's an article](https://blog.bam.tech/developer-news/should-you-use-enums-or-union-types-in-typescript) that weighs pros/cons of enum vs union type. 
2. Convention for enums is to use UpperCamelCase. But `ReturnType` is a built in type, so use a different name. `MethodReturnType` is the first thing that came to mind, but I'm not sure it's the best option. Naming things is hard.
3. `json` -> `parsed` just in case we ever get something where the response body is _not_ JSON, but the user wants us to handle it, rather than receive the raw response.

templates/operations.ts.hbs
```diff
@@ -121,7 +121,8 @@
         {{/if}}
       }>,
       rawResponse?: boolean,
-    ): Promise<Response | {{getReturnTypeFromOperation this}}> {
+      returnTypeParam?: returnType
```
wjhsf, line 124:
```suggestion
      returnType?: MethodReturnType
```
We already know that it's a parameter, because that's the context in which we are using it. 😉

templates/operations.ts.hbs
```diff
@@ -121,7 +121,8 @@
...
+    ): Promise<Response | Request | {{getReturnTypeFromOperation this}}> {
```
wjhsf, line 125: `Request` and `Response` are browser globals, but we also execute in a node environment and use the ones provided by `cross-fetch`. The differences between browser and node `fetch` implementations can be subtle, so make sure that the types used here are accurate.

templates/operations.ts.hbs
```diff
@@ -186,8 +187,11 @@
           method: "{{loud method}}"
         }
       );
+      if(returnTypeParam && returnTypeParam === returnType.REQUEST) {
```
wjhsf, line 190:
```suggestion
      if(returnTypeParam === returnType.REQUEST) {
```
Checking for truthiness is unnecessary if you're just doing an equality check. 😉

templates/operations.ts.hbs
```diff
@@ -186,8 +187,11 @@
...
+        return req
+      }
       const response = await fetch(req);
-      if (rawResponse) {
+      if (rawResponse || (returnTypeParam && returnTypeParam === returnType.RESPONSE)) {
```
wjhsf, line 194:
```suggestion
      if (returnTypeParam === returnType.RESPONSE || (rawResponse && returnTypeParam !== returnType.JSON)) {
```
If somebody does `someMethod({ returnTypeParam: returnType.JSON }, true)`, then `returnTypeParam` should take precedence over `rawResponse`.

templates/operations.ts.hbs
```diff
@@ -121,7 +121,8 @@
         {{/if}}
       }>,
       rawResponse?: boolean,
```
wjhsf, line 123: Make sure you add a deprecation notice to `rawResponse`. I know that `@deprecated` exists as a doc tag, but I don't know what scopes it can be applied to (i.e. can it be used for params or just for methods?).

wjhsf (changes requested):

PR #68, opened 2022-01-06
@w-8856911 Add SLAS helper

bredmond-sf:
There's a little more test coverage to go, but I'd otherwise like to get this PR rolling. This adds helpers for executing the Public Client SLAS login flows with PKCE. To see what this will look like in use, I've pushed a branch to the pwa-kit repo:
https://github.com/SalesforceCommerceCloud/pwa-kit/blob/br/slas-helper-demo/packages/pwa/app/commerce-api/auth.js

This PR also begins to separate README and Contributing docs


src/static/helpers/slasHelper.test.ts
```diff
@@ -0,0 +1,188 @@
+/*
+ * Copyright (c) 2021, salesforce.com, inc.
```
tarcang, line 2: Copyright 2022?

src/static/helpers/slasHelper.test.ts
```diff
@@ -0,0 +1,188 @@
...
+
+const authenticateCustomerMock = jest.fn(options => ({
+  url: 'https://localhost:3000/callback?usid=048adcfb-aa93-4978-be9e-09cb569fdcb9&code=J2lHm0cgXmnXpwDhjhLoyLJBoUAlBfxDY-AhjqGMC-o',
+}));
+
+const authorizeCustomerMock = jest.fn(options => ({
+  url: 'https://localhost:3000/callback?usid=048adcfb-aa93-4978-be9e-09cb569fdcb9&code=J2lHm0cgXmnXpwDhjhLoyLJBoUAlBfxDY-AhjqGMC-o',
+}));
+
+const getAccessTokenMock = jest.fn(options => ({access_token: 'access_token'})); // TODO: replace with proper access token
```
tarcang, line 19: Do we need to create stories for TODO's?

tarcang (dismissed):

src/static/helpers/slasHelper.test.ts
```diff
@@ -0,0 +1,188 @@
...
+  authorizeCustomer: authorizeCustomerMock,
+  getAccessToken: getAccessTokenMock,
+} as unknown as ShopperLogin<{
+  shortCode: string;
+  organizationId: string;
+  clientId: string;
+  siteId: string;
+}>;
+
+const mockParameters = {
```
wjhsf, line 39:
```suggestion
const parameters = {
```
These parameters aren't mocks, they're just regular parameters.

src/static/helpers/slasHelper.test.ts
```diff
@@ -0,0 +1,188 @@
...
+  });
+});
+
+describe('Authorize user', () => {
+  const expectedCode = 'J2lHm0cgXmnXpwDhjhLoyLJBoUAlBfxDY-AhjqGMC-o';
+
+  test('hits the authorize endpoint and recieves authorization code', async () => {
+    const authCode = (
+      await slasHelper.authorize(mockSlasClient, mockParameters)
+    ).code;
```
wjhsf, line 88:
```suggestion
    const { usid, code } = await slasHelper.authorize(mockSlasClient, mockParameters);
```
Does `usid` also need to be checked here?

src/static/helpers/slasHelper.test.ts
```diff
@@ -0,0 +1,188 @@
...
+    )?.parameters;
+
+    // required query parameters
+    // expect(parametersArg?.hint).toBe('guest'); TODO: revise this line
+    expect(parametersArg?.response_type).toBe('code');
+    expect(parametersArg).toHaveProperty('client_id');
+    expect(parametersArg).toHaveProperty('code_challenge');
+    expect(parametersArg).toHaveProperty('redirect_uri');
+
+    expect(authorizeCustomerMock).toBeCalled();
```
wjhsf, line 112: I'm not sure whether this test should focus on the method that is called or on the returned value. Depends on whether you consider the goal of the helper to be "makes and parses an HTTP request" or "calls an SDK method". If the former, the only the returned value should be checked, as the SDK method is a hidden implementation detail. If the latter, then using [`expect.objectContaining`](https://jestjs.io/docs/expect#expectobjectcontainingobject) provides a cleaner approach by allowing you to check against a subset of the called argument:
```suggestion
    expect(authorizeCustomerMock).toHaveBeenCalledWith({
      parameters: {
        response_type: 'code',
        // ...
      }
    });
```

src/static/helpers/slasHelper.test.ts
```diff
@@ -0,0 +1,188 @@
...
+  });
+
+  test('uses auth code and code verifier to generate JWT', async () => {
```
wjhsf, line 115: Same comment w.r.t what to test / `expect.objectContaining`.

src/static/helpers/slasHelper.test.ts
```diff
@@ -0,0 +1,188 @@
...
+    expect(tokenBodyArg.grant_type).toBe('authorization_code_pkce');
+    expect(tokenBodyArg).toHaveProperty('code_verifier');
+    expect(tokenBodyArg).toHaveProperty('code');
+    expect(tokenBodyArg).toHaveProperty('client_id');
+    expect(tokenBodyArg).toHaveProperty('redirect_uri');
+
+    expect(accessToken).toStrictEqual({access_token: 'access_token'});
+  });
+});
+describe('Registered B2C user flow', () => {
```
wjhsf, line 139: Same comment w.r.t what to test / `expect.objectContaining`. Also consider [`expect.stringMatching`](https://jestjs.io/docs/expect#expectstringmatchingstring--regexp)!

wjhsf (changes requested):

.depcheckrc.yml
```diff
@@ -1,9 +1,15 @@
...
   # Include an explanation why the package is ignored
-  - "@babel/*" # Used in babel.config.js and babel.config.jest.js
-  - "@types/react-dom" # Used in src/index.jsx
+  - '@babel/*' # Used in babel.config.js and babel.config.jest.js
+  - '@types/fs-extra'
+  - '@types/handlebars-helpers'
+  - '@types/node-fetch'
+  - '@types/react-dom' # Used in src/index.jsx
+  - fs-extra
+  - handlebars-helpers
```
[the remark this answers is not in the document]
bredmond-sf: Awesome, I'll incorporate that fix, thanks!

Contributing.md
```diff
@@ -0,0 +1,97 @@
...
+$ yarn install
+
+# Parse API files, render templates to src/lib folder and copy static files to src/lib
+$ yarn run renderTemplates
+
+# Launch sample application
+$ yarn start
+
+# Transpile and minify
+$ yarn build:lib
```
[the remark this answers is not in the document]
bredmond-sf: Appears to build the test app

src/static/helpers/slasHelper.test.ts
bredmond-sf: I prefer the test to focus on the latter "calls an SDK method"

src/static/helpers/slasHelper.test.ts
joeluong-sfcc: Nope, sorry those are just reminders for me. I wanted to push the unit tests I had already so I could get feed back as soon as I could

.depcheckrc.yml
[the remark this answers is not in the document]
bredmond-sf: done!

src/static/helpers/slasHelper.ts
```diff
@@ -0,0 +1,262 @@
...
+      code_challenge: codeChallenge,
+      ...(parameters.hint && {hint: parameters.hint}),
+      organizationId: slasClient.clientConfig.parameters.organizationId,
+      redirect_uri: parameters.redirectURI,
+      response_type: 'code',
+      ...(parameters.usid && {usid: parameters.usid}),
+    },
+  };
+
+  const response = await slasClient.authorizeCustomer(options, true);
```
johnboxall, line 116: To improve server-side peformance here, you want to split client and server side handling.

The SLAS API under `authorizeCustomer` returns an HTTP 303 redirect to the callback URL in the `Location` header.

Server-side, you'd rather not follow the redirect (skip making another HTTP request) as you can dig the USID and code values you need out of `Location` header.

Client-side, you must follow the redirect because the fetch spec says that redirects are opaque and you cannot access the header: https://fetch.spec.whatwg.org/#atomic-http-redirect-handling

I've tried to do this poorly in the PWA Kit:

https://github.com/SalesforceCommerceCloud/pwa-kit/pull/254/files
https://github.com/SalesforceCommerceCloud/pwa-kit/blob/f300639e26453e17c755f208c93ec5d68f788390/packages/pwa/app/commerce-api/auth.js#L272-L290

bredmond-sf: I'll need to play with the server-side a little which I'll try to do today. On the client side the behavior is:
1. Fetch call is made with default 'follow'
2. Server returns 303
3. A call is made to the redirect url by the browser (visible in network tab)
4. We ignore that and instead access the `url` property of the initial 303 response from the server and grab the code and usid
You can add the 'manual' option to avoid the browser making the redirect call, but then you aren't able to read anything from the response including that `url` property. I can see from the link you put that we should be able to use the 'manual' option on the server and still get the url so I'll try that, thanks!

package.json
```diff
@@ -92,7 +92,8 @@
     "**/yargs": "<17"
   },
   "dependencies": {
-    "cross-fetch": "^3.1.4"
+    "cross-fetch": "^3.1.4",
+    "nanoid": "^3.1.30"
```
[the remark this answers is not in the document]
bredmond-sf: https://gus.lightning.force.com/lightning/r/ADM_Third_Party_Software__c/a0qEE0000001QEzYAM/view

src/static/helpers/slasHelper.ts
```diff
@@ -0,0 +1,258 @@
...
+  const urlSafe = (input: string) =>
+    input.replace(/\+/g, '-').replace(/\//g, '_').replace(/=/g, '');
+
+  let challenge = '';
+  try {
+    if (window.crypto.subtle.digest !== undefined) {
+      const encoder = new TextEncoder();
+      const data = encoder.encode(codeVerifier);
+      const digest = await window.crypto.subtle.digest('SHA-256', data);
+      const base64Digest = btoa(String.fromCharCode(...new Uint8Array(digest)));
```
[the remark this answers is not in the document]
bredmond-sf: this https://stackoverflow.com/questions/9267899/arraybuffer-to-base64-encoded-string

src/static/helpers/slasHelper.ts
```diff
@@ -0,0 +1,258 @@
...
+
+  const options = {
+    parameters: {
+      client_id: slasClient.clientConfig.parameters.clientId,
+      code_challenge: codeChallenge,
+      ...(parameters.hint && {hint: parameters.hint}),
+      organizationId: slasClient.clientConfig.parameters.organizationId,
+      redirect_uri: parameters.redirectURI,
+      response_type: 'code',
+      ...(parameters.usid && {usid: parameters.usid}),
```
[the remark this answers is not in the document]
bredmond-sf: Some of this may point to things that need tweaking in the generation at some point, but I get not assignable for `authorizeCustomer` without explicitly setting `clientId` and `organizationId`. Passing `usid` and `hint` were causing problems at runtime because the query params were being sent empty and the service was complaining (I know that shouldn't be the case looking at the code, but that's what was happening.)

src/static/helpers/slasHelper.ts
bredmond-sf: Done! conditionally setting 'manual' for serverside calls

src/static/helpers/slasHelper.ts
```diff
@@ -0,0 +1,272 @@
...
+ */
+export const generateCodeChallenge = async (
+  codeVerifier: string
+): Promise<string> => {
+  const urlSafe = (input: string) =>
+    input.replace(/\+/g, '-').replace(/\//g, '_').replace(/=/g, '');
+
+  let challenge = '';
+  try {
+    // Cannot test browser functions
```
wjhsf, line 56: You could do `globalThis.window = ...`, but that's probably not worth it.

src/static/helpers/slasHelper.ts
```diff
@@ -0,0 +1,258 @@
...
+ * @param slasRefreshToken
+ * @returns
+ */
+export function refreshToken(
+  slasClient: ShopperLogin<{
+    shortCode: string;
+    organizationId: string;
+    clientId: string;
+  }>,
+  parameters: {refreshToken: string}
```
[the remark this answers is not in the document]
bredmond-sf: I'd agree, but I'm preferring to leave it because 'refresh' being a noun and verb here means you have a clash between the function name and parameter name. Since keeping `parameters` is consist with the other functions *and* prevents the name clash, it gets my vote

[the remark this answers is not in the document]
wjhsf: That's a good point, though, that you need a refresh token to refresh the access token. Could the method name be more clear?

src/static/helpers/slasHelper.ts
```diff
@@ -0,0 +1,258 @@
...
+ * @returns code verifier
+ */
+export const createCodeVerifier = (): string => nanoid(128);
+
+/**
+ * Encodes a code verifier to a code challenge to send to the authorization endpoint
+ * @param codeVerifier random string to use as a code verifier
+ * @returns code challenge
+ */
+export const generateCodeChallenge = async (
```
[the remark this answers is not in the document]
bredmond-sf: done!

src/static/helpers/slasHelper.ts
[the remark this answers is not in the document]
bredmond-sf: I'm struggling to come up with it myself, but is `refreshAccessToken` clearer maybe?

src/static/helpers/slasHelper.test.ts
```diff
@@ -0,0 +1,258 @@
...
+    authenticateCustomer: authenticateCustomerMock,
+    authorizeCustomer: authorizeCustomerMock,
+    getAccessToken: getAccessTokenMock,
+    logoutCustomer: logoutCustomerMock,
+  } as unknown as ShopperLogin<{
+    shortCode: string;
+    organizationId: string;
+    clientId: string;
+    siteId: string;
+  }>);
```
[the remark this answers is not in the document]
joeluong-sfcc: decided to go with `jest.clearAllMocks()` to accomplish this

src/static/helpers/slasHelper.test.ts
```diff
@@ -0,0 +1,263 @@
+/**
+ * @jest-environment node
+ */
```
wjhsf, line 3: Why is this added?

src/static/helpers/slasHelper.test.ts
```diff
@@ -0,0 +1,263 @@
...
+    clientId: string;
+    siteId: string;
+  }>);
+
+beforeEach(() => {
+  jest.clearAllMocks();
+});
+
+describe('Create code verifier', () => {
+  test('creates 128 URL safe string', () => {
```
wjhsf, line 75:
```suggestion
  test('creates 128-character URL-safe string', () => {
```
I hope it doesn't create 128 URLs 🥴

wjhsf (approved):

README.md
```diff
@@ -2,106 +2,119 @@
...
-}
+  proxy: 'https://localhost:3000', // Routes API calls through a proxy when set
+  headers: {},
+  parameters: {
+    clientId: '<your-client-id>',
+    organizationId: '<your-org-id>',
+    shortCode: '<your-short-code>',
+    siteId: '<your-site-id>',
+  },
+  throwOnBadResponse: true,
```
tarcang, line 48: We want to throw on bad responses as a default configuration?

tarcang (approved):

package.json
```diff
@@ -91,15 +91,18 @@
     "**/@oclif/command": "<=1.8.3",
```
wjhsf, line 91:
```suggestion
    "**/@npmcli/fs": "<1.1.0",
    "**/@oclif/command": "<=1.8.3",
```
This will fix the failing node 10 build.

src/static/helpers/slasHelper.test.ts
joeluong-sfcc: We changed some things so the default test environment is jsdom and certain tests fail because the test environment is not node

README.md
bredmond-sf: We don't have it as default in the SDK because it would be a breaking change. Ideally, it would have been the default behavior from the start, so I'm adding it in this example since it is what nearly everyone wants.

wjhsf (approved):

wjhsf (approved):

tarcang (approved):

PR #75, opened 2022-02-21
W-10519931: SLAS Login Helper should not request redirect_uri when running server-side

joeluong-sfcc:


src/static/helpers/slasHelper.ts
```diff
@@ -111,10 +111,7 @@ export async function authorize(
   // set manual redirect on server since node allows access to the location
   // header and it skips the extra call. In the browser, only the default
   // follow setting allows us to get the url.
-  const fetchOptions = {
-    ...slasClient.clientConfig.fetchOptions,
-    ...(!onClient && {redirect: 'manual'}),
-  };
+  const redirect: RequestRedirect = 'manual';
```
wjhsf, line 114: This sets `redirect` to `"manual"` regardless of environment, but the comment seems to suggest that the setting in browser must be `"follow"`. What is the intended behavior in browser?

src/static/helpers/slasHelper.ts
```diff
@@ -126,12 +123,12 @@ export async function authorize(
...
       ...(parameters.usid && {usid: parameters.usid}),
     },
-    fetchOptions,
+    redirect,
   };
 
   const response = await slasClient.authorizeCustomer(options, true);
 
-  const redirectUrl = response.url || response.headers?.get('location');
+  const redirectUrl = response.headers?.get('location') || response.url;
```
wjhsf, line 131: Why the switch?

templates/operations.ts.hbs
```diff
@@ -180,6 +183,7 @@
         url.toString(),
         {
           ...this.clientConfig.fetchOptions,
+          redirect: options?.redirect || this.clientConfig.fetchOptions.redirect || 'follow',
```
wjhsf, line 186: Why does a new top-level option need to be added?

wjhsf (changes requested):

src/static/helpers/slasHelper.ts
```diff
@@ -126,12 +123,12 @@ export async function authorize(
       response_type: 'code',
       ...(parameters.usid && {usid: parameters.usid}),
     },
-    fetchOptions,
+    redirect,
```
bredmond-sf, line 126: `redirect` needs to be a field within `fetchOptions`

src/static/helpers/slasHelper.ts
bredmond-sf: Exactly. In node, you can set `manual` and still get a readable Response that includes the location header. In the browser, setting it to `manual` gives you an unreadable Response. You need to leave it on `follow` and then you can use the `url` field. https://stackoverflow.com/questions/39735496/redirect-after-a-fetch-post-call

src/static/helpers/slasHelper.ts
bredmond-sf: Is this maybe the only change that needs to be made is this one line?

src/static/helpers/slasHelper.ts
```diff
@@ -111,7 +111,8 @@ export async function authorize(
   // set manual redirect on server since node allows access to the location
   // header and it skips the extra call. In the browser, only the default
   // follow setting allows us to get the url.
-  const fetchOptions = {
+  const slasClientCopy = new ShopperLogin(slasClient.clientConfig);
```
unandyala, line 114: please add a comment on why we need another copy of slasclient

src/static/helpers/slasHelper.ts
```diff
@@ -108,10 +108,13 @@ export async function authorize(
...
+  // Create a copy to override specific fetchOptions
+  const slasClientCopy = new ShopperLogin(slasClient.clientConfig);
+
   // set manual redirect on server since node allows access to the location
   // header and it skips the extra call. In the browser, only the default
   // follow setting allows us to get the url.
-  const fetchOptions = {
+  slasClientCopy.clientConfig.fetchOptions = {
     ...slasClient.clientConfig.fetchOptions,
     ...(!onClient && {redirect: 'manual'}),
```
wjhsf, line 119:
```suggestion
    redirect: onClient ? 'follow' : 'manual',
```
This is safer, because it guarantees the right redirect setting in both node and browser, rather than just in node. (Also I think it looks cleaner.)

Another option, while I'm here, is to use spread in the constructor so you only need one statement:
```typescript
  const slasClientCopy = new ShopperLogin({
    ...slasClient.clientConfig,
    fetchOptions: {
      ...slasClient.clientConfig.fetchOptions,
      redirect: onClient ? 'follow' : 'manual'
  });
```
Not sure which is cleaner.

joeluong-sfcc: Leaning towards keeping the code as multiple statements. Seems a little easier to read especially with the comments

wjhsf (approved):

tarcang (approved):

PR #85, opened 2022-04-14
Don't set user-agent in browser because it will fail.

wjhsf:
It fails noisily, and that's scary! 🤫

### DIFF
--- a/src/static/helpers/environment.ts
+++ b/src/static/helpers/environment.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022, salesforce.com, inc.
+ * Copyright (c) 2022, Salesforce, Inc.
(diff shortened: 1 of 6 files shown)


src/static/helpers/environment.ts
```diff
@@ -0,0 +1,13 @@
+/*
+ * Copyright (c) 2022, salesforce.com, inc.
```
unandyala, line 2: I believe the copyright now should have Salesforce, Inc.

unandyala (changes requested):

tarcang (approved):

joeluong-sfcc (approved):

unandyala (approved):

PR #86, opened 2022-04-22
Misc. patches for release

wjhsf:
* Get above code coverage threshold
* Fix circular dependency
* Make rollup quiet

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -56,7 +56,8 @@ const config = outputs.map(({file, format}) => ({
     includePaths({
       include: {},
       paths: ['src'],
-      external: Object.keys(pkg.dependencies),
+      // crypto is a node built-in, but rollup doesn't seem to know that?
+      external: ['crypto', ...Object.keys(pkg.dependencies)],
       extensions: ['.js', '.json', '.html'],
     }),
     stylelint({
--- a/src/static/clientConfig.test.ts
+++ b/src/static/clientConfig.test.ts
@@ -63,6 +63,15 @@ describe('ClientConfig constructor', () => {
       originalConfig.fetchOptions.timeout
     );
   });
+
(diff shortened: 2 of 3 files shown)


tarcang (approved):

rollup.config.js
```diff
@@ -56,7 +56,8 @@ const config = outputs.map(({file, format}) => ({
     includePaths({
       include: {},
       paths: ['src'],
-      external: Object.keys(pkg.dependencies),
+      // crypto is a node built-in, but rollup doesn't seem to know that?
+      external: ['crypto', ...Object.keys(pkg.dependencies)],
```
unandyala, line 60: do we need to add "crypto" to globals? Did you test the SDK with this change?

wjhsf: Without this change: this is the output when building:
```
(!) Unresolved dependencies
https://rollupjs.org/guide/en/#warning-treating-module-as-external-dependency
crypto (imported by src/lib/helpers/slasHelper.ts)
```
From [that link](https://rollupjs.org/guide/en/#warning-treating-module-as-external-dependency):
> Rollup will only resolve relative module IDs by default. This means that an import statement like this…
> ```js
> import moment from 'moment';
> ```
> …won't result in moment being included in your bundle – instead, it will be an external dependency that is required at runtime. If that's what you want, you can suppress this warning with the external option, which makes your intentions explicit:
> ```js
> // rollup.config.js
> export default {
>   entry: 'src/index.js',
>   dest: 'bundle.js',
>   format: 'cjs',
>   external: ['moment'] // <-- suppresses the warning
> };
> ```

I don't know why rollup doesn't recognize that `crypto` is a [node.js built-in](https://nodejs.org/docs/latest/api/crypto.html), but this change just suppresses that line of log output.

PR #90, opened 2022-06-09
 loginRegisteredUserB2C method to not request redirect_uri

tarcang:
As a developer using the isomorphic SDK, I would like the loginRegisteredUserB2C method to not request redirect_uri when running server-side so that my code does not fail if there is not a valid server running on redirect_uri.

AC:

Request is made with manual redirect
Errors are handled in loginRegisteredUserB2C() and the other helpers

### DIFF
--- a/src/static/helpers/slasHelper.ts
+++ b/src/static/helpers/slasHelper.ts
@@ -241,9 +241,17 @@ export async function loginRegisteredUserB2C(
     },
   };
 
-  const response = await slasClientCopy.authenticateCustomer(options, true);
+  const response = await slasClientCopy
+    .authenticateCustomer(options, true)
+    .catch((error: Error) => {
+      throw new Error(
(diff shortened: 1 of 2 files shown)


src/static/helpers/slasHelper.ts
```diff
@@ -230,9 +244,15 @@ export async function loginRegisteredUserB2C(
     },
   };
 
-  const response = await slasClient.authenticateCustomer(options, true);
+  const response = await slasClientCopy
+    .authenticateCustomer(options, true)
+    .catch((error: Error) => {
```
gkhunger-sf, line 249: We seem to be just wrapping one error into another here.  We want to check for any client or server errors, for which we'll need to look at the status inside the response.

gkhunger-sf (approved):

unandyala (approved):

joeluong-sfcc (approved):

PR #91, opened 2022-06-22
Added a test to check the urlencoded body transition

tarcang:
The story seems to be already implemented by @wjhsf a couple of months ago with these lines in operations.ts.hbs: 
`      const headers: Record<string, string> = {
        {{#if (isRequestWithPayload request)}}
        "Content-Type": "{{{getMediaTypeFromRequest request}}}",
        {{/if}}
        ...this.clientConfig.headers,
        ...options?.headers
      };
`
In this PR I have added a test that checks if the content-type is urlencoded, the body should be urlencoded.

### DIFF
--- a/src/test/requests.test.ts
+++ b/src/test/requests.test.ts
@@ -1,0 +1,55 @@
+/*
+ * Copyright (c) 2022, salesforce.com, inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: BSD-3-Clause
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+import nock from 'nock';
+import {ShopperLogin} from '../lib';
+import ClientConfig from '../static/clientConfig';
+
+const config: ClientConfig<any> = {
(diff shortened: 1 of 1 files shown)


src/test/requests.test.ts
```diff
@@ -0,0 +1,55 @@
...
+ * Copyright (c) 2022, salesforce.com, inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: BSD-3-Clause
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+import nock from 'nock';
+import {ShopperLogin} from '../lib';
+import ClientConfig from '../static/clientConfig';
+
+const config: ClientConfig<any> = {
```
joeluong-sfcc, line 11: Lint code pretty please :)

tarcang: 👌

joeluong-sfcc (dismissed):

gkhunger-sf (commented):
Could you add a test for the json payload use case as well please?

gkhunger-sf (approved):

PR #97, opened 2022-07-26
W-11358406: Fix logout function in SLAS helpers

joeluong-sfcc:
Small PR that fixes the `logout` function and updates the `authorize` function to use the `ResponseError` class for error handling.


src/static/helpers/slasHelper.ts
```diff
@@ -306,9 +307,15 @@ export function logout(
     clientId: string;
     siteId: string;
   }>,
-  parameters: {refreshToken: string}
+  parameters: {
+    accessToken: string;
```
wjhsf, line 311: This violates the established pattern of `parameters` being "things in the URL". I don't have a better solution, though.

unandyala (approved):

gkhunger-sf (approved):

PR #100, opened 2022-08-11
Fix SLAS callback error handling

kevinxh:
Hey team,

I believe there is a bug with the SLAS helpers, specifically the ones that deals with redirect callbacks: `authorize` and `loginRegisteredUserB2C`.

Currently these two helpers will throw an error if the response status is not 303. However, on the browser environment, we always follow the redirect and by following the redirect, the response status will not be 303, but the response from the actual callback endpoint, which is likely a 200 or something. Screenshot below shows that the callback 200 causes the helpers to throw.

Problematic code:
```js
const response = await slasClientCopy.authenticateCustomer(options, true);

// when the redirect is followed on the browser, status will not be 303
// and the if statement will always evaluate to true and throws error
if (response.status !== 303) {
  throw new ResponseError(response);
}
```

Screen Shot 2022-08-11 at 12 30 31 AM

Secondly, there are two scenarios when we need to throw errors:
1. when SLAS API returns error, which the status code is >= 400.
2. when the redirect callback has an error query string `/callback?error=invalid_request` <- this is not handled!

The PR fixes the above two issues.

I've tested the solution on my PWA testing project and it seems the issue is gone with the changes in this PR.

### DIFF
--- a/src/static/helpers/slasHelper.test.ts
+++ b/src/static/helpers/slasHelper.test.ts
@@ -168,14 +168,18 @@ describe('Authorize user', () => {
     expect(authResponse).toStrictEqual(expectedAuthResponse);
   });
 
-  test('throws error on non 303 response', async () => {
+  test('throws error on error response', async () => {
     const mockSlasClient = createMockSlasClient();
     const {shortCode, organizationId} = mockSlasClient.clientConfig.parameters;
 
     nock(`https://${shortCode}.api.commercecloud.salesforce.com`)
       .get(`/shopper/auth/v1/organizations/${organizationId}/oauth2/authorize`)
       .query(true)
-      .reply(400, {response_body: 'response_body'}, {location: ''});
+      .reply(303, undefined, {
+        Location: '/callback?error=invalid_request',
+      })
+      .get('/callback?error=invalid_request')
+      .reply(200);
 
     await expect(
       slasHelper.authorize(mockSlasClient, codeVerifier, parameters)
--- a/src/static/helpers/slasHelper.test.ts
+++ b/src/static/helpers/slasHelper.test.ts
@@ -168,7 +168,7 @@ describe('Authorize user', () => {
     expect(authResponse).toStrictEqual(expectedAuthResponse);
   });
 
-  test('throws error on error response', async () => {
+  test('throws error on SLAS callback error response', async () => {
     const mockSlasClient = createMockSlasClient();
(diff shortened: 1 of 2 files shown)


src/static/helpers/slasHelper.test.ts
```diff
@@ -168,21 +168,39 @@ describe('Authorize user', () => {
...
 
     nock(`https://${shortCode}.api.commercecloud.salesforce.com`)
       .get(`/shopper/auth/v1/organizations/${organizationId}/oauth2/authorize`)
       .query(true)
-      .reply(400, {response_body: 'response_body'}, {location: ''});
+      .reply(303, undefined, {
+        Location: '/callback?error=invalid_request',
+      })
+      .get('/callback?error=invalid_request')
+      .reply(200);
```
kevinxh, line 182: This test case mimics the actual followed redirect on a browser environment.

src/static/helpers/slasHelper.ts
```diff
@@ -131,16 +131,18 @@ export async function authorize(
...
-  const redirectUrl = response.headers?.get('location') || response.url;
+  const redirectUrlString = response.headers?.get('location') || response.url;
+  const redirectUrl = new URL(redirectUrlString);
+  const searchParams = Object.fromEntries(redirectUrl.searchParams.entries());
 
   // url is a read only property we unfortunately cannot mock out using nock
   // meaning redirectUrl will not have a falsy value for unit tests
   /* istanbul ignore next */
-  if (response.status !== 303 || !redirectUrl) {
+  if (response.status >= 400 || searchParams.error) {
```
joeluong-sfcc, line 141: Hey Kevin everything looks good so I'll go ahead and approve it, but is there a specific reason you error check against the `searchParams`? Just asking for clarity!

joeluong-sfcc (approved):

src/static/helpers/slasHelper.ts
kevinxh: Hey Joel,

The redirect callback is sometimes called with the error parameter `/callback?error=invalid_request`. We need to throw an error when this happens.

bendvc (approved):

src/static/helpers/slasHelper.ts
joeluong-sfcc: AH I apologize I just read your description more throughly, thanks for clarifying!

gkhunger-sf (approved):

PR #112, opened 2023-02-01
fix: replace anypoint exchange url with a mock url

joeluong-sfcc:


tarcang (approved):
Looks good if it passes the pipelines!

PR #115, opened 2023-03-08
Tg/disable cookies by default

tarcang:
- Added credentials:'omit' in the request object
- Calling native fetch if browser

Right now I have implemented the fix by basically by setting a if statement to check if the caller is browser. This might be better implemented with https://www.typescriptlang.org/docs/handbook/2/objects.html#intersection-types 
I am looking into that now

### DIFF
--- a/src/static/helpers/environment.ts
+++ b/src/static/helpers/environment.ts
@@ -11,3 +11,6 @@ export const isNode =
   typeof process === 'object' &&
   typeof process.versions === 'object' &&
   typeof process.versions.node === 'string';
+
+export const hasFetchAvailable = 
+  typeof fetch === 'function' && typeof Request === 'function';
--- a/templates/client.ts.hbs
+++ b/templates/client.ts.hbs
@@ -1,8 +1,8 @@
-import { fetch as fetchCross, Request }  from "cross-fetch";
+import { fetch as crossFetch, Request as CrossRequest }  from "cross-fetch";
 
 import ClientConfig, { ClientConfigInit } from "./clientConfig";
 // Must not import from ./helpers/index to avoid circular dependency via ShopperLogin
-import { isBrowser } from "./helpers/environment";
+import { isBrowser, hasFetchAvailable } from "./helpers/environment";
 import type {
   BaseUriParameters,
   CompositeParameters,
--- a/templates/client.ts.hbs
+++ b/templates/client.ts.hbs
@@ -1,4 +1,4 @@
-import { fetch, Request }  from "cross-fetch";
+import { fetch as fetchCross, Request }  from "cross-fetch";
 
 import ClientConfig, { ClientConfigInit } from "./clientConfig";
 // Must not import from ./helpers/index to avoid circular dependency via ShopperLogin
--- a/templates/operations.ts.hbs
+++ b/templates/operations.ts.hbs
@@ -174,21 +174,41 @@
         ...options?.headers
       };
 
-      if (!isBrowser) {
-        // Browsers forbid setting a custom user-agent header
-        headers[USER_AGENT_HEADER] = [headers[USER_AGENT_HEADER], USER_AGENT_VALUE].join(" ");
+      let response;
+      if(isBrowser && hasFetchAvailable){
+        const req = new Request(
+          url.toString(),
+          {
+            ...this.clientConfig.fetchOptions,
+            {{#if (isRequestWithPayload request)}}body: this.clientConfig.transformRequest(options.body, headers),{{/if}}
+            headers,
+            credentials: 'omit',
+            method: "{{loud method}}"
+          }
+        );
+
+        response = await fetch(req);
+      }
+      else {
+          if (!isBrowser) {
+            // Browsers forbid setting a custom user-agent header
+            headers[USER_AGENT_HEADER] = [headers[USER_AGENT_HEADER], USER_AGENT_VALUE].join(" ");
+          }
+
+          const req = new CrossRequest(
+          url.toString(),
+          {
+            ...this.clientConfig.fetchOptions,
+            {{#if (isRequestWithPayload request)}}body: this.clientConfig.transformRequest(options.body, headers),{{/if}}
+            headers,
+            credentials: 'omit',
+            method: "{{loud method}}"
+          }
+        );
+
+        response = await crossFetch(req);
       }
 
-      const req = new Request(
-        url.toString(),
-        {
-          ...this.clientConfig.fetchOptions,
-          {{#if (isRequestWithPayload request)}}body: this.clientConfig.transformRequest(options.body, headers),{{/if}}
-          headers,
-          method: "{{loud method}}"
-        }
-      );
-      const response = isBrowser && typeof fetch === 'function' ? await fetch(req) : await fetchCross(req);
       if (rawResponse) {
         return response;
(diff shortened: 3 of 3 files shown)


johnboxall:
@tarcang – one opportunity we have with this change is to completely remove the `fetch` polyfill from the browser.

There are two at least two reasons why this would be a good change:

1. Having the fallback behaviour introduces inconsistent behaviour. This behaviour won't work in browsers that don't support `fetch`.
2. In 2023, all modern browsers have `fetch`: https://caniuse.com/?search=fetch – removing the polyfill reduced the size of this download, speeding things up.

The last holdout was IE11 which is now dead: https://learn.microsoft.com/en-us/lifecycle/products/internet-explorer-11

templates/client.ts.hbs
```diff
@@ -12,6 +12,9 @@ import ResponseError from "./responseError";
 import TemplateURL from "./templateUrl";
 import { USER_AGENT_HEADER, USER_AGENT_VALUE } from "./version";
 
+// The goal is to run native fetch in the 'real' browser. The reason isNode is also checked so to not use the fetch in jest-dom (nock tests) or other browser mocking libraries.
+const [runFetchStrategy, ValidRequest] = isBrowser && hasFetchAvailable && !isNode ? [fetch, Request] : [crossFetch, CrossRequest];
```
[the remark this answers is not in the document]
tarcang: Yes, I did verify manually in the browser. 

This is the current state of the fetch call. Notice the red request (right now it is failing because of wrong credentials but notice the type and initiator)
image

If the cross fetch called, it would say this:
Screenshot 2023-03-10 at 10 53 46 AM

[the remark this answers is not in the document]
joeluong-sfcc: Awesome! Were you able to verify it in a node environment?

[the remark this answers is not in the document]
tarcang: Not manually, no. I thought the tests already cover the node environment (and the react test is also in node environment because it mocks the browser)

templates/client.ts.hbs
```diff
@@ -12,6 +10,12 @@ import ResponseError from "./responseError";
 import TemplateURL from "./templateUrl";
 import { USER_AGENT_HEADER, USER_AGENT_VALUE } from "./version";
 
+if(isNode || !isBrowser || !hasFetchAvailable) {
+    const nodeFetch = require('node-fetch');
+    global.fetch = nodeFetch;
+    global.Request = nodeFetch.Request;
```
johnboxall, line 16: @tarcang – these lines may have unintended side effects.

If you import `commerce-sdk-isomorphic`, it will overwrite `global.fetch` with the `node-fetch` implementation ... 

In Node 18+, `global.fetch` is something different, with a different interface.

Is there anyway to set `fetch` such that we don't overwrite the global?

templates/client.ts.hbs
```diff
@@ -12,6 +10,12 @@ import ResponseError from "./responseError";
 import TemplateURL from "./templateUrl";
 import { USER_AGENT_HEADER, USER_AGENT_VALUE } from "./version";
 
+if(isNode || !isBrowser || !hasFetchAvailable) {
```
johnboxall, line 13: I'm not sure I entirely understand this conditional.

I think, if we're in a Node.js environment, we want to load `node-fetch`. If you're not in a Node environment, and you don't have `fetch`, well, the jig is up. We could consider throwing an error?

p-biyani (approved):

templates/client.ts.hbs
tarcang: I have edited the logic to throw an error if we are not in node but fetch is not defined.

templates/client.ts.hbs
tarcang: I was able to do it :)

p-biyani (dismissed):

johnboxall (dismissed):
🚢 🚢 🚢 

jboxall_Create_a_Studio_Ghibli-inspired_anime_illustration_of_a_a54df80b-7dde-4e4f-a67e-18692fd935a7

wjhsf (changes requested):
Things that are broken: `node-fetch` is missing, and accessing `global` in a browser context will throw an error.

wjhsf:
Oh, also I _think_ it's already configured, but definitely validate that `node-fetch` is _not_ included in the bundle, otherwise the file shipped to browsers will include code that's never used.

src/static/clientConfig.ts
```diff
@@ -64,6 +70,7 @@ export default class ClientConfig<Params extends BaseUriParameters>
       throw new Error('Missing required parameter: shortCode');
     }
     this.fetchOptions = {...config.fetchOptions};
+    this.fetchOptions.credentials = config.fetchOptions?.credentials || 'omit';
```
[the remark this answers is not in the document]
tarcang: I see, spread does overwrite properties if added later. 👍

bredmond-sf (dismissed):

src/static/helpers/environment.test.ts
```diff
@@ -0,0 +1,19 @@
...
+
+import nodeFetch from 'node-fetch';
+import {isNode, hasFetchAvailable, fetch} from './environment';
+
+/* Just testing the node environment, testing the browser environment is too complex within the test cases. */
+
+describe('Fetch', () => {
+  test('Runs node-fetch if node', () => {
+    expect(isNode).toBe(true);
+    expect(hasFetchAvailable).toBe(false);
```
wjhsf, line 16: This'll break on node 18, but that's a future problem!

src/static/helpers/environment.ts
```diff
@@ -1,3 +1,11 @@
+/*
+ * Copyright (c) 2023, Salesforce, Inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: BSD-3-Clause
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+import {FetchFunction} from 'lib/clientConfig';
```
wjhsf, line 7:
```suggestion
import type {FetchFunction} from '../clientConfig';
```
1.`import type` tells typescript that there's no JS with side effects (e.g. modifying globals) in the imported files, so typescript can remove the import when compiling.
2. Imports from `lib` break for me locally 😐

wjhsf (approved):

package.json
```diff
@@ -96,8 +96,8 @@
     "depcheck/@babel/parser": "7.16.4"
   },
   "dependencies": {
-    "cross-fetch": "^3.1.5",
-    "nanoid": "^3.3.4"
+    "nanoid": "^3.3.4",
+    "node-fetch": "2"
```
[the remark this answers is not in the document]
tarcang: We cannot use 'require' with node-fetch 3 (since it is a full ESM module). Node-fetch 2 is still widely supported and new versions are still coming up.

src/static/helpers/environment.ts
tarcang: Oh, good point! If this works better, I can merge this in.

src/static/helpers/environment.test.ts
tarcang: I can mine as well remove that line, no need for future headaches.

joeluong-sfcc (approved):

bredmond-sf (approved):

PR #116, opened 2023-03-15
Tg/experimental remove cross fetch

tarcang:
Trying to remove "cross-fetch" so we can optimize the bundle size.


tarcang:
Covered by #115

PR #120, opened 2023-04-07
Handle Shopper Context title change @W-12969581

tarcang:
**Background**

- The release process is blocked because of the title change in Shopper Context. The type name and the class name (which was derived from the title in raml) caused this issue.

**How does these changes fix this?**

- In these changes, we overwrite the `name` property of the `ShopperContext` API model returned by the RAML Toolkit. To make the name change work with the template generation, we had to pass the `updateName` parameter for the init method in `ApiModel` [(documentation)](https://github.com/SalesforceCommerceCloud/raml-toolkit/blob/b1f3a6b9f24e3f91a3af54eb2523139891f0d6ac/src/generate/apiModel.ts#L105) (for some reason `ApiMetadata` object doesn't have this parameter in its `init` method) as false. This is a crucial part in this fix since our name overwrite wouldn't matter, raml-toolkit would change the name property on its own.

- Also some test cases are updated after the breaking changes in the API.

### DIFF
--- a/scripts/utils.ts
+++ b/scripts/utils.ts
@@ -69,9 +69,9 @@ function addTemplates(apis: ApiMetadata, outputBasePath: string): ApiMetadata {
   return apis;
 }
 
+// Editing the name of the Shopper Context API Model so our name is used for generating class name.
+// This is hard-coded for now but in the future if we handle the case where type name and title clash better, this can be removed.
 function overwriteShopperContextName(apis: ApiMetadata): void {
-  // Editing the name of the Shopper Context API Model so our name is used for generating class name.
-  // This is hard-coded for now but in the future if we handle the case where type name and title clash better, this can be removed.
   const shopperContextApi = apis.children.find(
     api => api.name.original === 'shopper-context'
   );
--- a/scripts/utils.ts
+++ b/scripts/utils.ts
@@ -69,6 +69,18 @@ function addTemplates(apis: ApiMetadata, outputBasePath: string): ApiMetadata {
   return apis;
 }
 
+function overwriteShopperContextName(apis: ApiMetadata): void {
+  // Editing the name of the Shopper Context API Model so our name is used for generating class name.
+  // This is hard-coded for now but in the future if we handle the case where type name and title clash better, this can be removed.
+  const shopperContextApi = apis.children.find(
+    api => api.name.original === 'shopper-context'
+  );
+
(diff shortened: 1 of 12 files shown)


scripts/utils.ts
```diff
@@ -84,7 +86,20 @@ export async function setupApis(
   // is necessary for generating the SDK (as part of the user agent header).
   // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
   apis.metadata.sdkVersion = await readJsonSync(PACKAGE_JSON).version;
-  await apis.init();
+
+  // Editing the name of the Shopper Context API Model so our name is used for generating class name.
```
unandyala, line 90: Can you add a story in the backlog to fix the name collision and revert this hack. And the gus ID in the comments

scripts/utils.ts
```diff
@@ -84,7 +86,20 @@ export async function setupApis(
...
+  // This is hard-coded for now but in the future if we handle the case where type name and title clash better, this can be removed.
+  const shopperContextApi = apis.children.find(
```
unandyala, line 92: can we extract this code into a specific function or method (updateShopperContextName). That way it is clean and easy to understand when we want to remove the hack

scripts/utils.ts
```diff
@@ -84,7 +86,20 @@ export async function setupApis(
...
+    api => api.name.original === 'shopper-context'
+  );
+
+  if (shopperContextApi) {
+    shopperContextApi.name = new Name('shopper-contexts');
+  }
+
+  // We are calling the init for children (which will call loadModel) since we can pass the updateName parameter as false (default was true).
+  // With this way the ApiModel doesn't change its name again from the loaded api model, which will use the raml title field.
+  await Promise.all(apis.children.map(api => (api as ApiModel).init(false)));
```
unandyala, line 102: what  happens if you don't reload the model?

unandyala: you are calling` .init(false) `for all the APIs while we need it only shopper-context. Can we avoid that as there is a possibility to break customer implementations

tarcang: I can totally try to call with (false) just for Shopper Context. I don't think this current version can break the changes now because the titles are consistent with the download names. I do agree it is safer to call this just for Shopper Context.

bredmond-sf (approved):

PR #131, opened 2023-08-10
Fix Annoying things with this Package. 

bertybot:
## Why this Change?

This package is currently misconfigured. Running it through [publint](https://publint.dev/commerce-sdk-isomorphic@1.10.1) it has a few errors that need to be addressed.

The main problem being that the ESM configuration is currently misconfigured causing this package to fail to build in ESM centric environments like Vite.

- node-fetch import is also always a require which causes ESM builds to fail.

There are also a few other annoying things with this package that should be addressed. 
- React is a peer dependency even though this package compiles to pure JS and works perfectly fine without React.
- There is no option to supply your own fetch function basically forcing you to use 2 outdated fetch functions even if you have no need for them.

## What was done
- Correctly configured package.json
    -  added `exports` field
    -  removed deprecated `module` field
    -  added `packageManager` field so people can more easily contribute to the project
- Added optional  `fetchFunction` property to config so that you can supply your own fetch function if you don't like the 2 supplied
- Fixed `node-fetch` import
- updated `node-fetch` 

## Conclusion

These changes most likely constitute a breaking change to users on older versions of Node. However, currently these problems currently make this package completely unusable to people on newer versions or using ESM so I think that the juice is worth the squeeze.

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,8 +11,14 @@
     "url": "git+https://github.com/SalesforceCommerceCloud/commerce-sdk-isomorphic.git"
   },
   "license": "BSD-3-Clause",
-  "main": "lib/index.cjs.js",
-  "module": "lib/index.esm.js",
+  "exports": {
+    ".": {
+      "require": "./lib/index.cjs",
+      "import": "./lib/index.mjs",
+      "types": "./lib/index.d.ts"
+    }
+  },
+  "main": "./lib/index.cjs",
   "style": "lib/default.css",
   "files": [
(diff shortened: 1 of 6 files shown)


wjhsf:
Hi @bertybot, thanks for creating this PR! There are a lot of good suggestions in here, and we'd like to include most of them. However, now that node 18 has native fetch support, we will be able to rely on native `fetch` in all environments, and not need any logic for providing a custom implementation.

PR #135, opened 2023-09-15
Fix expand query param @W-14129454@

joeluong-sfcc:
This PR changes the logic in how arrays as query parameters are encoded:
- Before: repeated format where `{ a: [1, 2] } => '?a=1&a=2'`
- After: comma separated format where `{ a: [1, 2] } => '?a=1,2'`

This is because SCAPI expects arrays to be comma separated. The exception to this rule is the refine query parameter as SCAPI expects this to be repeated:
- `{ refine: [1, 2] } => '?refine=1&refine=2'`

This PR also fixes the inconsistencies between the typing of the `expand` query parameter between `shopper-search` and `shopper-products` where before one was of type array and the other was of type string. They are now both of type array.


bfeister (commented):
@joeluong-sfcc can you add a description to this PR por favor?

src/static/templateUrl.ts
```diff
@@ -50,17 +50,22 @@ export default class TemplateURL extends URL {
...
+   * convention where \{ refine: ["1", "2"] \} becomes "?refine=1&refine=2"
    */
   addQueryParams(queryParams?: QueryParameters): void {
     if (queryParams) {
       Object.keys(queryParams).forEach(key => {
         const param = queryParams[key];
         if (Array.isArray(param)) {
-          for (let i = 0; i < param.length; i += 1) {
-            this.searchParams.append(key, String(param[i]));
+          if (key === 'refine') {
```
bfeister, line 63: I think this is OK, but if we ever allow anything else to behave the same way as refine we might want an "allow-list" approach. That being said, I think doing so might be premature optimization, since our best guess is that we won't be doing this with any other key

bfeister (approved):
Thanks for the quick turnaround on this, I know it was unplanned but very important to help our sister teams 💪 🏅

wjhsf (approved):

PR #136, opened 2023-09-15
Release v1.10.3 @W-14129454@

joeluong-sfcc:
This PR releases `v1.10.3` which fixes the expand query parameter in shopper-products to match shopper-search, so both are arrays. This PR also updates SLAS, which now has the `refresh_token_expires_in` field in `TokenResponse` and removes `retrieveCredQualityUserInfo` endpoint which was removed from SLAS.


docs/modules/version.html
```diff
@@ -95,7 +95,7 @@ <h3><span class="tsd-flag ts-flagConst">Const</span> USER_<wbr>AGENT_<wbr>HEADER
 				<section class="tsd-panel tsd-member tsd-kind-variable tsd-parent-kind-module">
 					<a name="user_agent_value" class="tsd-anchor"></a>
 					<h3><span class="tsd-flag ts-flagConst">Const</span> USER_<wbr>AGENT_<wbr>VALUE</h3>
-					<div class="tsd-signature tsd-kind-icon">USER_<wbr>AGENT_<wbr>VALUE<span class="tsd-signature-symbol">:</span> <span class="tsd-signature-type">"commerce-sdk-isomorphic@1.10.1"</span><span class="tsd-signature-symbol"> = &quot;commerce-sdk-isomorphic@1.10.1&quot;</span></div>
+					<div class="tsd-signature tsd-kind-icon">USER_<wbr>AGENT_<wbr>VALUE<span class="tsd-signature-symbol">:</span> <span class="tsd-signature-type">"commerce-sdk-isomorphic@1.10.2"</span><span class="tsd-signature-symbol"> = &quot;commerce-sdk-isomorphic@1.10.2&quot;</span></div>
```
vmarta, line 98: 👀 should this be `1.10.3`?

joeluong-sfcc: hmm yeah, that's a great question, this html file links here: https://salesforcecommercecloud.github.io/commerce-sdk-isomorphic/modules/version.html

I checked the past couple of releases and it seems like the version always lags behind. Not entirely sure why, but we don't usually edit these HTML files directly as they get automatically generated from a script. I think it should be fine to leave it as is since it doesn't have any functional impact

vmarta (approved):

PR #138, opened 2023-11-28
DO NOT MERGE - Add a separate function for calling getAccessToken

vcua-mobify:
DO NOT MERGE

To support both public and private clients as well as 3rd party IDPs, we could export a function that allows for more flexibility in calling SLAS /token endpoint.

Thinking is that if a refresh token is included, we can assume the user wants a refresh token login. This grant type supersedes all other types of request.

If no refresh token, check if the user provides a code verifier. If yes, assume the user is using a SLAS public client (only public clients need to provide a code verifier as the [PKCE](https://auth0.com/docs/get-started/authentication-and-authorization-flow/authorization-code-flow-with-proof-key-for-code-exchange-pkce))

If no code verifier, we can assume the user is using a SLAS private client. If a code is still provided, set grant_type to `authorization_code`. This is used for registered user federated login.

If no code is provided, assume we are logging in a guest user via SLAS private client and set grant_type to `client_credentials`

### DIFF
--- a/src/static/helpers/slasHelper.ts
+++ b/src/static/helpers/slasHelper.ts
@@ -148,6 +148,70 @@ export async function authorize(
 }
 
 /**
+ * Wrapper for the /token endpoint. Depending on the parameters provided, will set up different grant types.
+ * @param slasClient a configured instance of the ShopperLogin SDK client
+ */
+export async function getAccessToken(
+  slasClient: ShopperLogin<{
+  shortCode: string;
+  organizationId: string;
+  clientId: string;
+  siteId: string;
+}>,
+parameters: {
+  redirectURI: string;
+  refreshToken?: string;
+  codeVerifier?: string;
+  code?: string;
+  usid?: string;
+},
+credentials?: {
+  client_id: string;
+  client_secret: string;
+}
+): Promise<TokenResponse> {
+
+  let tokenHeader = {}
+  if (credentials) {
+    const authorization = `Basic ${stringToBase64(
+      `${credentials.client_id}:${credentials.client_secret}`
+    )}`;
+
+    tokenHeader = {
+        Authorization: authorization
+    }
+  }
+
+  const grantType = parameters.refreshToken ? 'refresh_token'
+    : parameters.codeVerifier ? 'authorization_code_pkce'
+    : parameters.code ? 'authorization_code'
+    : 'client_credentials'
+
+  const tokenBody: TokenRequest = {
+    client_id: slasClient.clientConfig.parameters.clientId,
+    channel_id: slasClient.clientConfig.parameters.siteId,
+    grant_type: grantType,
+    redirect_uri: parameters.redirectURI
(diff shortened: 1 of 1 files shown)


src/static/helpers/slasHelper.ts
```diff
@@ -188,6 +252,29 @@ export async function loginGuestUser(
...
+  },
+  credentials?: {
+    client_id: string;
+    client_secret: string;
+  }
+  ): Promise<TokenResponse> {
+
+  return await getAccessToken(slasClient, parameters, credentials);
+}
+
```
vcua-mobify, line 277: commerce-sdk-isomorphic assumes we are using a SLAS public client

When using a SLAS private client, we want to skip the call to `/authorize`. We can immediately call the `/token` endpoint

src/static/helpers/slasHelper.ts
```diff
@@ -147,6 +147,70 @@ export async function authorize(
   return {url: redirectUrlString, ...getCodeAndUsidFromUrl(redirectUrlString)};
 }
 
+/**
+ * Wrapper for the /token endpoint. Depending on the parameters provided, will set up different grant types.
+ * @param slasClient a configured instance of the ShopperLogin SDK client
+ */
+export async function getAccessToken(
```
vcua-mobify, line 154: I added a catch-all wrapper for the getAccessToken endpoint that can handle refresh tokens, pkce (public client) logins, and client-credentials (private client) logins.

Later on, we could refactor the other helper methods to call this rather than setting up the `/token` call for a specific case in many places.

src/static/helpers/slasHelper.ts
```diff
@@ -147,6 +147,70 @@ export async function authorize(
...
+
+    tokenHeader = {
+        Authorization: authorization
+    }
+  }
+
+  const grantType = parameters.refreshToken ? 'refresh_token'
+    : parameters.codeVerifier ? 'authorization_code_pkce'
+    : parameters.code ? 'authorization_code'
+    : 'client_credentials'
```
vcua-mobify, line 188: The parameters we provide to `/token` affect the `grant_type` and what the endpoint ultimately does. 

I think the order of this conditional is the correct priority order for how to handle scenarios where multiple parameters corresponding to different grant_types are provided.

src/static/helpers/slasHelper.ts
```diff
@@ -147,6 +147,70 @@ export async function authorize(
...
+  redirectURI: string;
+  refreshToken?: string;
+  codeVerifier?: string;
+  code?: string;
+  usid?: string;
+},
+credentials?: {
+  client_id: string;
+  client_secret: string;
+}
```
vcua-mobify, line 171: In the PWA Kit, we're using a proxy to intercept the call to SLAS and inject the client secret there. So PWA Kit does not set this arg.

However, this arg is here to enable non-PWA Kit projects that may want to use a private client.

PR #140, opened 2024-01-20
@W-14758586: Support slas client

alexvuong:
This PR adds support for SLAS [private](https://developer.salesforce.com/docs/commerce/commerce-api/guide/slas-private-client.html) client without making a breaking change.

How to test 
1. `git clone git@github.com:SalesforceCommerceCloud/commerce-sdk-isomorphic.git`
2. `git checkout support-slas-client`
3. yarn install
4. npm run renderTemplates && yarn build:lib
5. Create a node project in a separate directory
```
mkdir testSlasPrivateClient
cd testSlasPrivateClient
npm init -y
touch index.js
```
6. modify package.json in testSlasPrivateClient directory to point commerce-sdk-isomorphic to local directory in dependencies section and add "type": "module":
```
  "type": "module",
  "dependencies": {
    "commerce-sdk-isomorphic": "<insert_path_to_local_commerce-sdk-isomorphic>"
  }
```
7. npm install
8. Set up SLAS private client via API for the instance that you are testing on https://kv7kzm78.api.commercecloud.salesforce.com/shopper/auth-admin/v1/sso/login
9. Add test code to index.js and replace client creds
```
const { helpers, ShopperLogin, ShopperSearch } = pkg;

// demo client credentials, if you have access to your own please replace them below.
const CLIENT_ID = "<CLIENT_ID>";
const ORG_ID = "<ORG_ID>";
const SHORT_CODE = "<SHORT_CODE>";
const SITE_ID = "<SITE_ID>";
const CLIENT_SECRET = '<SLAS PRIVATE SECRET HERE>' // SLAS PRIVATE SECRET HERE
//  // EDIT HERE: Fill in with shopper credentials. Examples on how to register a shopper can be found in 04-register-shopper.ts
 const shopper = {
   username: <username>,
   password: <password>, // do not store password as plaintext. Store it in a secure location.
 };
// must be registered in SLAS. On server, redirectURI is never called
const redirectURI = "http://localhost:3000/callback";

const clientConfig = {
    parameters: {
        clientId: CLIENT_ID,
        organizationId: ORG_ID,
        shortCode: SHORT_CODE,
        siteId: SITE_ID,
    },
};

const slasClient = new ShopperLogin(clientConfig);

// GUEST LOGIN
const guestTokenResponse = await helpers
    .loginGuestUserPrivate(slasClient, { redirectURI }, {clientSecret: CLIENT_SECRET})
    .then((guestTokenResponse) => {
        console.log("Guest Token Response: ", guestTokenResponse);
        return guestTokenResponse;
    })
    .catch((error) =>
        console.log("Error fetching token for guest login: ", error)
    );
// uncommnent the function you want to test
//  // REGISTERED B2C USER LOGIN
//  await helpers
//    .loginRegisteredUserB2C(
//      slasClient,
//      { username: shopper.username, password: shopper.password, clientSecret: CLIENT_SECRET },
//      { redirectURI }
//    )
//    .then((registeredUserTokenResponse) => {
//      console.log(
//        "Registered User Token Response: ",
//        registeredUserTokenResponse
//      );
//      return registeredUserTokenResponse;
//    })
//    .catch((error) =>
//      console.log("Error fetching token for registered user login: ", error)
//    );

// REFRESH TOKEN
//  const refreshTokenResponse = await helpers
//    .refreshAccessToken(slasClient, {
//      refreshToken: guestTokenResponse.refresh_token,
//    }, {clientSecret: CLIENT_SECRET})
//    .then((refreshTokenResponse) => {
//      console.log("Refresh Token Response: ", refreshTokenResponse);
//      return refreshTokenResponse;
//    })
//    .catch((error) => console.log("Error refreshing token: ", error));

// using JWT to make SCAPI API for additional check that private slas is working as expected
const shopperSearch = new ShopperSearch({
    ...clientConfig,
    headers: {authorization: `Bearer ${guestTokenResponse.access_token}`},
});

const searchResult = await shopperSearch.productSearch({
    parameters: {q: 'shirt'},
}).then((result) => {
    console.log("RESULT", result)
});

console.log(searchResult);
```
10. node index.js > out.txt
11. Observe out.txt and that the access_token is successfully retrieved with slas private secret and can be used to make SCAPI APIs

### DIFF
--- a/src/static/helpers/slasHelper.test.ts
+++ b/src/static/helpers/slasHelper.test.ts
@@ -22,6 +22,12 @@ const credentials = {
   password: 'shopper_password',
 };
 
+const credentialsPrivate = {
+  username: 'shopper_user_id',
+  password: 'shopper_password',
+  clientSecret: 'slas_private_secret',
+};
+
 const expectedTokenResponse: TokenResponse = {
   access_token: 'access_token',
   id_token: 'id_token',
@@ -204,18 +210,18 @@ test('throws error on 400 response', async () => {
 });
 
 describe('Guest user flow', () => {
-  const expectedTokenBody = {
-    body: {
-      client_id: 'client_id',
-      channel_id: 'site_id',
-      code: 'J2lHm0cgXmnXpwDhjhLoyLJBoUAlBfxDY-AhjqGMC-o',
-      code_verifier: expect.stringMatching(/./) as string,
-      grant_type: 'authorization_code_pkce',
-      redirect_uri: 'redirect_uri',
-      usid: '048adcfb-aa93-4978-be9e-09cb569fdcb9',
-    },
-  };
   test('retrieves usid and code from location header and generates an access token', async () => {
+    const expectedTokenBody = {
+      body: {
+        client_id: 'client_id',
+        channel_id: 'site_id',
+        code: 'J2lHm0cgXmnXpwDhjhLoyLJBoUAlBfxDY-AhjqGMC-o',
+        code_verifier: expect.stringMatching(/./) as string,
+        grant_type: 'authorization_code_pkce',
+        redirect_uri: 'redirect_uri',
+        usid: '048adcfb-aa93-4978-be9e-09cb569fdcb9',
+      },
+    };
     const mockSlasClient = createMockSlasClient();
     const {shortCode, organizationId} = mockSlasClient.clientConfig.parameters;
 
@@ -229,6 +235,30 @@ describe('Guest user flow', () => {
       parameters
     );
     expect(getAccessTokenMock).toBeCalledWith(expectedTokenBody);
+    expect(accessToken).toBe(expectedTokenResponse);
+  });
+
+  test('generates an access token using slas private client', async () => {
+    const mockSlasClient = createMockSlasClient();
+
+    const accessToken = await slasHelper.loginGuestUserPrivate(
+      mockSlasClient,
+      parameters,
+      credentialsPrivate
+    );
+
+    const expectedReqOptions = {
+      headers: {
+        Authorization: `Basic ${stringToBase64(
+          `client_id:${credentialsPrivate.clientSecret}`
+        )}`,
+      },
+      body: {
+        grant_type: 'client_credentials',
+        usid: 'usid',
+      },
+    };
+    expect(getAccessTokenMock).toBeCalledWith(expectedReqOptions);
     expect(accessToken).toBe(expectedTokenResponse);
   });
 });
--- a/src/static/helpers/slasHelper.test.ts
+++ b/src/static/helpers/slasHelper.test.ts
@@ -276,7 +276,7 @@ describe('Registered B2C user flow', () => {
     },
   };
 
-  test('uses code challenge and authorization header to generate auth code', async () => {
+  test('uses code challenge and authorization header to generate auth code with slas public client', async () => {
     // slasClient is copied and tries to make an actual API call
     const mockSlasClient = createMockSlasClient();
     const {shortCode, organizationId} = mockSlasClient.clientConfig.parameters;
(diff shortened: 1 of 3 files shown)


src/static/helpers/slasHelper.ts
```diff
@@ -16,6 +16,9 @@ import {
 } from '../../lib/shopperLogin';
 import ResponseError from '../responseError';
 
+const warningSlasSecretMsg =
+  'This function can run on client-side. You are potentially exposing SLAS secret on browser. Make sure to keep it safe and secure!';
+
```
joeluong-sfcc, line 21: I think we should also include a similar warning in the README with the addition of a private SLAS client helper method

src/static/helpers/slasHelper.ts
```diff
@@ -147,6 +150,55 @@ export async function authorize(
...
+    `${slasClient.clientConfig.parameters.clientId}:${credentials.clientSecret}`
+  )}`;
+
+  const options = {
+    headers: {
+      Authorization: authorization,
+    },
+    body: {
+      grant_type: 'client_credentials',
+      ...(parameters.usid && {usid: parameters.usid}),
```
joeluong-sfcc, line 195: We accept `redirectURI` but never pass it onto the request, it should be a part of the body: https://developer.salesforce.com/docs/commerce/commerce-api/references/shopper-login?meta=getAccessToken

src/static/helpers/slasHelper.ts
```diff
@@ -280,6 +363,9 @@ export async function loginRegisteredUserB2C(
  * @param slasClient a configured instance of the ShopperLogin SDK client.
  * @param parameters - parameters to pass in the API calls.
  * @param parameters.refreshToken - a valid refresh token to exchange for a new access token (and refresh token).
+ * @param parameters.refreshToken - a valid refresh token to exchange for a new access token (and refresh token).
```
joeluong-sfcc, line 366:
```suggestion
```

param comment is duplicated from line above

src/static/helpers/slasHelper.test.ts
```diff
@@ -340,23 +404,68 @@ describe('Registered B2C user flow', () => {
...
+      parameters,
+      {
+        clientSecret: credentialsPrivate.clientSecret,
+      }
+    );
+    expect(getAccessTokenMock).toBeCalledWith(expectedReqOpts);
+    expect(token).toStrictEqual(expectedTokenResponse);
+  });
+
+  test('refreshes the token warn users about using slas private client on', () => {
```
joeluong-sfcc, line 447: This seems like a duplicate to the test above it, is the plan to modify this test to check the warning?

src/static/helpers/slasHelper.ts
```diff
@@ -260,7 +314,36 @@ export async function loginRegisteredUserB2C(
...
+        grant_type: 'authorization_code_pkce',
+        code_verifier: codeVerifier,
+        code: authResponse.code,
+        client_id: slasClient.clientConfig.parameters.clientId,
+        redirect_uri: parameters.redirectURI,
+        usid: authResponse.usid,
+      },
+    };
+    return slasClient.getAccessToken(optionsToken);
+  }
```
vcua-mobify, line 344: I think we can reuse the same optionsToken as the public client. When logging in a registered user, only difference in requirements between a public and private client is whether we need to include the auth header with the client Id and secret.

We still want the other parameters like siteId to be set in the body.

vcua-mobify: Another way to think about it is that since the grant type here in both cases is authorization_code_pkce, the parameters in the body should be identical.

src/static/helpers/slasHelper.ts
alexvuong: I read the private flow again, we don't need to pass redirectURI for private flow. uuisd and grant_type should suffice https://developer.salesforce.com/docs/commerce/commerce-api/guide/slas-private-client.html

src/static/helpers/slasHelper.test.ts
alexvuong: good catch. It has been removed

src/static/helpers/slasHelper.ts
alexvuong: I left a note in readme and added a section for private slas client

joeluong-sfcc (dismissed):

src/static/helpers/slasHelper.ts
```diff
@@ -298,6 +373,25 @@ export function refreshAccessToken(
     channel_id: slasClient.clientConfig.parameters.siteId,
   };
 
+  if (credentials && credentials.clientSecret) {
+    /* istanbul ignore next */
+    if (isBrowser) {
+      // we intentionally have warning here
+      // eslint-disable-next-line
+      console.warn(warningSlasSecretMsg);
+    }
```
vcua-mobify, line 382: Is there a way to suppress this warning if we are using a placeholder value for the client secret or will it always appear as long as this function runs client side.

alexvuong: it will always appear as long as the func runs on client side. I don't use any place holder value here in this library since I find it unnecessary.

joeluong-sfcc (dismissed):

vcua-mobify (commented):
I played around with the test script and everything seems to be good to go.

One thing I am curious about the warning. Since it always appears, is there a possibility of the warning appearing on a production site and an end shopper seeing the warning?

joeluong-sfcc (dismissed):

vcua-mobify (approved):
Thanks for looking into the warning @alexvuong and for applying the feedback.

PR #142, opened 2024-01-29
Release v1.11.0 @W-14879795@

joeluong-sfcc:
This PR prepares the release for v1.11.0. The main changes are:
- Support for SLAS private client with helper functions
- Support for custom query parameters

The SDK has also been updated to explicitly throw errors when an invalid parameter is passed in instead of filtering the query parameter out without notifying the user: https://github.com/SalesforceCommerceCloud/commerce-sdk-isomorphic/pull/142/files#diff-efb305ab11f90c680288cff20266dc8f6a3cbb40189be6aba09a9268081da7fbR163

For ease of review, the `docs/` directory is auto generated, and the `apis/` directory changes are pulled from anypoint exchange


alexvuong (dismissed):

bfeister (approved):
So the RAML here does all the work expressing the new option? I suppose it's mainly the Typescript types in `commerce-sdk-react` that will require more non-generated code changes

templates/operations.ts.hbs
```diff
@@ -159,6 +159,8 @@
       Object.keys(optionParams).forEach((key) => {
         if(key.startsWith('c_') && optionParams[key as keyof typeof optionParams] !== undefined) {
           queryParams[key as keyof typeof queryParams] = optionParams[key as keyof typeof optionParams]
+        } else if(!queryParams.hasOwnProperty(key) && !pathParams.hasOwnProperty(key)) {
+          throw new Error(`Invalid Parameter: ${key}`)
```
bendvc, line 163: This would be considered a breaking change. Do we need to be doing this? 
Also, it wouldn't be in alignment with the comment in the docs of this other pr [here](https://github.com/SalesforceCommerceCloud/commerce-sdk/pull/398/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R230)

bendvc (changes requested):
You made a breaking change, but only did a feature bump. Shall we discuss?

bendvc (dismissed):

joeluong-sfcc:
> So the RAML here does all the work expressing the new option? I suppose it's mainly the Typescript types in `commerce-sdk-react` that will require more non-generated code changes

@bfeister By new option are you referring to the custom query parameters, SLAS private client helpers, or something else? If you're referring to the custom query parameters, it's the handlebar templates provide the type for custom query parameters while the RAML provides all the other query parameter types

bendvc (approved):

alexvuong (approved):

PR #148, opened 2024-04-10
@W-15470872 Release 1.13.1

alexvuong:
This Release fixes the missing parameters (offset and limit) happening in v1.13.0


apis/shopper-baskets/shopper-baskets-description.md
```diff
@@ -7,9 +7,9 @@ Temporary baskets can perform calculations to generate totals, line items, promo
...
 
-To create a basket, start with the [Create basket](https://developer.salesforce.com/docs/commerce/commerce-api/references/shopper-baskets?meta=createBasket) endpoint. The endpoint creates the basket in the B2C Commerce system and returns a JSON representation of the basket with a `basketId` property.
+To create a basket, start with the [Create basket](https://developer.salesforce.com/docs/commerce/commerce-api/references/shopper-baskets-v2?meta=createBasket) endpoint. The endpoint creates the basket in the B2C Commerce system and returns a JSON representation of the basket with a `basketId` property.
 
-If you provide the JSON for a prepopulated basket to the [Create basket](https://developer.salesforce.com/docs/commerce/commerce-api/references/shopper-baskets?meta=createBasket) endpoint, you can create a basket using a single API request.
+If you provide the JSON for a prepopulated basket to the [Create basket](https://developer.salesforce.com/docs/commerce/commerce-api/references/shopper-baskets-v2?meta=createBasket) endpoint, you can create a basket using a single API request.
```
bendvc, line 12: What happening here? Why are we referencing the v2 basket endpoint and replacing the v1 comments?

bendvc (changes requested):
Taking a quick look at the size of this change, there looks to be a lot more than just the offset and other param fix in here. I don't think this is a point release.

apis/shopper-baskets/shopper-baskets-description.md
alexvuong: This is not new, I think we have been updated for basket in v1.11.0. See [here](https://github.com/SalesforceCommerceCloud/commerce-sdk-isomorphic/pull/142/files#diff-36b34c6728cac9504992b8e4174fa0ad3b4a0c730732f317d3a0a98c2aaedfaaR5) 🤔

joeluong-sfcc: I think I missed this when I released `v1.11.0` as I wasn't fully aware that there was the introduction of the v2 of shopper baskets at the time, that's my bad, but we'll need to figure out how to support 2 different versions of the same API in the SDK.

Good news is that I don't believe anything has been modified in the RAML to constitute a breaking change in the SDK, the only thing that was modified was the addition of a query parameter, `temporary`, for `createBasket` that only works on `v2` and some updates to error responses

This isn't relevant for this patch release, but we should probably start omitting `shopper-baskets` changes until we figure out a way to support both v1 and v2 in `shopper-baskets`

bendvc (approved):

PR #149, opened 2024-04-16
Feat: support custom APIs @W-15111169@

joeluong-sfcc:
This PR adds support for custom APIs by exposing a helper function called `callCustomEndpoint` that allows users to call their custom API following common SDK patterns. This PR also pulls out some of the reused logic in `operations.ts.hbs` for fetching data and puts into a helper function called `doFetch`.

Some features of the `callCustomEndpoint` helper:
1. Defaults `apiVersion` to `v1` if not provided
2. Defaults `content-type` header to `application/json` if not provided 

Below is a sample script I used for testing the `callCustomEndpoint` helper function. Reach out to me for instructions on how to run this with the proper credentials.

```javascript
const { helpers } = pkg;

// replace client credentials
const CLIENT_ID = "clientId";
const ORG_ID = "orgId";
const SHORT_CODE = "shortCode";
const SITE_ID = "siteId";

const clientConfig = {
  parameters: {
    clientId: CLIENT_ID,
    organizationId: ORG_ID,
    shortCode: SHORT_CODE,
    siteId: SITE_ID,
    endpointPath: 'greeting',
    apiName: 'e2e-tests',
    apiVersion: 'v1',
  },
  // proxy: 'http://localhost:8888/mobify/proxy/api'
  // baseUri: 'https://{shortCode}.alternativeBaseUri.com/custom/{apiName}/{apiVersion}'
};

const access_token = "<INSERT_ACCESS_TOKEN>"

let response = await helpers.callCustomEndpoint({
    method: 'GET',
    headers: {
        'Content-Type': 'application/json',
        authorization: `Bearer ${access_token}`
    }
}, clientConfig, false)

console.log('RESPONSE: ', response)

```

Resulting fetch request URLs:
1. No proxy/baseUri: `https://shortcode.api.commercecloud.salesforce.com/custom/e2e-tests/v1/organizations/orgId/greeting?siteId=siteId`
4. With Proxy: `http://localhost:8888/mobify/proxy/api/custom/e2e-tests/v1/organizations/orgId/greeting?siteId=siteId`
5. With baseUri: `https://shortcode.alternativebaseuri.com/custom/e2e-tests/v1/organizations/orgId/greeting?siteId=siteId`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # CHANGELOG
 
+## v1.14.0-dev
+
+#### Enchancements
+
+- Add support for custom APIs [#149](https://github.com/SalesforceCommerceCloud/commerce-sdk-isomorphic/pull/149)
+
 ## v1.13.1
 
 #### Bug fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 #### Enchancements
 
-- Add support for custom APIs with `callCustomEndpoint` helper function [#149](https://github.com/SalesforceCommerceCloud/commerce-sdk-isomorphic/pull/149)
+- Add helper function `callCustomEndpoint` to call [Custom APIs](https://developer.salesforce.com/docs/commerce/commerce-api/guide/custom-apis.html) - [#149](https://github.com/SalesforceCommerceCloud/commerce-sdk-isomorphic/pull/149)
 
 ## v1.13.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 #### Enchancements
 
-- Add support for custom APIs [#149](https://github.com/SalesforceCommerceCloud/commerce-sdk-isomorphic/pull/149)
+- Add support for custom APIs with `callCustomEndpoint` helper function [#149](https://github.com/SalesforceCommerceCloud/commerce-sdk-isomorphic/pull/149)
 
 ## v1.13.1
 
--- a/README.md
+++ b/README.md
@@ -151,18 +151,13 @@ Example usage:
 import pkg from 'commerce-sdk-isomorphic';
 const { helpers } = pkg;
 
-const CLIENT_ID = "<your-client-id>";
-const ORG_ID = "<your-org-id>";
-const SHORT_CODE = "<your-short-code>";
-const SITE_ID = "<your-site-id>";
-
 // client configuration parameters
 const clientConfigExample = {
   parameters: {
-    clientId: CLIENT_ID,
-    organizationId: ORG_ID,
-    shortCode: SHORT_CODE,
-    siteId: SITE_ID,
+    clientId: "<your-client-id>",
+    organizationId: "<your-org-id>",
+    shortCode: "<your-short-code>",
+    siteId: "<your-site-id>",
   }
 };
 
@@ -202,7 +197,6 @@ let postResponse = await helpers.callCustomEndpoint({
       queryParameter: 'queryParameter1',
     },
     headers: {
-      'Content-Type': 'application/json',
       authorization: `Bearer ${access_token}`
     },
     customApiPathParameters: customApiArgs,
--- a/README.md
+++ b/README.md
@@ -151,7 +151,6 @@ Example usage:
 import pkg from 'commerce-sdk-isomorphic';
 const { helpers } = pkg;
 
-// client configuration parameters
 const clientConfigExample = {
   parameters: {
     clientId: "<your-client-id>",
--- a/README.md
+++ b/README.md
@@ -143,7 +143,7 @@ Invalid query parameters that are not a part of the API and do not follow the `c
 
 ### Custom APIs
 
-The SDK supports calling custom APIs with a helper function, `callCustomEndpoint`.
+The SDK supports calling [custom APIs](https://developer.salesforce.com/docs/commerce/commerce-api/guide/custom-apis.html) with a helper function, `callCustomEndpoint`.
 
 Example usage:
 
@@ -157,26 +157,48 @@ const SHORT_CODE = "<your-short-code>";
 const SITE_ID = "<your-site-id>";
 
 // client configuration parameters
-const clientConfig = {
+const clientConfigExample = {
   parameters: {
     clientId: CLIENT_ID,
     organizationId: ORG_ID,
     shortCode: SHORT_CODE,
     siteId: SITE_ID,
-    // Custom API path parameters
-    endpointPath: 'customers',
-    apiName: 'loyalty-info',
-    apiVersion: 'v1', // defaults to v1 if not provided
-  },
+  }
 };
 
-// Flag to retrieve raw response or data from helper function
-const rawResponse = false;
+// Required params: apiName, endpointPath, shortCode, organizaitonId
+// Required path params can be passed into:
+// options.customApiPathParameters or clientConfig.parameters
+const customApiArgs = { 
+  endpointPath: 'customers',
+  apiName: 'loyalty-info',
+  apiVersion: 'v1', // defaults to v1 if not provided
+}
+
 const accessToken = '<INSERT ACCESS TOKEN HERE>';
 
-let response = await helpers.callCustomEndpoint(
-  {
+let getResponse = await helpers.callCustomEndpoint({
+  options: {
     method: 'GET',
+    parameters: {
+      queryParameter: 'queryParameter1',
+      siteId: SITE_ID,
+    },
+    headers: {
+      // Content-Type is defaulted to application/json if not provided
+      'Content-Type': 'application/json',
+      authorization: `Bearer ${access_token}`
+    },
+    customApiPathParameters: customApiArgs
+  }, 
+  clientConfig: clientConfigExample,
+  // Flag to retrieve raw response or data from helper function
+  rawResponse: false, 
+})
+
+let postResponse = await helpers.callCustomEndpoint({
+  options: {
+    method: 'POST',
     parameters: {
       queryParameter: 'queryParameter1',
       siteId: SITE_ID,
@@ -184,18 +206,22 @@ let response = await helpers.callCustomEndpoint(
     headers: {
       'Content-Type': 'application/json',
       authorization: `Bearer ${access_token}`
-    }
+    },
+    customApiPathParameters: customApiArgs,
+    body: JSON.stringify({ data: 'data' })
   }, 
-  clientConfig,
-  rawResponse
-)
+  clientConfig: clientConfigExample,
+  // Flag to retrieve raw response or data from helper function
+  rawResponse: false, 
+})
 
-console.log('RESPONSE: ', response)
+console.log('get response: ', getResponse)
+console.log('post response: ', postResponse)
 ```
 
 For more documentation about this helper function, please refer to the [commerce-sdk-isomorphic docs](https://salesforcecommercecloud.github.io/commerce-sdk-isomorphic/modules/helpers.html).
 
-For more information about custom APIs, please refer to the [Salesforce Developer Docs](https://developer.salesforce.com/docs/commerce/commerce-api/guide/custom-apis.html?q=custom+API)
+For more information about custom APIs, please refer to the [Salesforce Developer Docs](https://developer.salesforce.com/docs/commerce/commerce-api/guide/custom-apis.html)
 
 ## License Information
 
--- a/README.md
+++ b/README.md
@@ -157,7 +157,11 @@ const clientConfigExample = {
     organizationId: "<your-org-id>",
     shortCode: "<your-short-code>",
     siteId: "<your-site-id>",
-  }
+  },
+  // If not provided, it'll use the default production URI:
+  // 'https://{shortCode}.api.commercecloud.salesforce.com/custom/{apiName}/{apiVersion}'
+  // path parameters should be wrapped in curly braces like the default production URI
+  baseUri: "<your-base-uri>"
 };
 
 // Required params: apiName, endpointPath, shortCode, organizaitonId
--- a/README.md
+++ b/README.md
@@ -141,6 +141,62 @@ const searchResult = await shopperSearch.productSearch({
 
 Invalid query parameters that are not a part of the API and do not follow the `c_` custom query parameter convention will be filtered from the request and a warning will be displayed.
 
+### Custom APIs
+
+The SDK supports calling custom APIs with a helper function, `callCustomEndpoint`.
+
+Example usage:
+
+```javascript
+import pkg from 'commerce-sdk-isomorphic';
+const { helpers } = pkg;
+
+const CLIENT_ID = "<your-client-id>";
+const ORG_ID = "<your-org-id>";
+const SHORT_CODE = "<your-short-code>";
+const SITE_ID = "<your-site-id>";
+
+// client configuration parameters
+const clientConfig = {
+  parameters: {
+    clientId: CLIENT_ID,
+    organizationId: ORG_ID,
+    shortCode: SHORT_CODE,
+    siteId: SITE_ID,
+    // Custom API path parameters
+    endpointPath: 'customers',
+    apiName: 'loyalty-info',
+    apiVersion: 'v1', // defaults to v1 if not provided
+  },
+};
+
+// Flag to retrieve raw response or data from helper function
+const rawResponse = false;
+const accessToken = '<INSERT ACCESS TOKEN HERE>';
+
+let response = await helpers.callCustomEndpoint(
+  {
+    method: 'GET',
+    parameters: {
+      queryParameter: 'queryParameter1',
+      siteId: SITE_ID,
+    },
+    headers: {
+      'Content-Type': 'application/json',
+      authorization: `Bearer ${access_token}`
+    }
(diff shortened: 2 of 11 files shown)


src/static/helpers/fetchHelper.ts
```diff
@@ -0,0 +1,68 @@
...
+ * @param options.body? - Body that is used for the request
+ * @param clientConfig? - Client Configuration object used by the SDK with properties that can affect the fetch call
+ * @param clientConfig.headers? - Additional headers that are added to the request. Authorization header should be in this argument or in the options?.headers. options?.headers will override any duplicate properties.
+ * @param clientConfig.fetchOptions? - fetchOptions that are passed onto the fetch request
+ * @param clientConfig.throwOnBadResponse? - flag that when set true will throw a response error if the fetch request fails
+ * @param rawResponse? - Flag to return the raw response from the fetch call. True for raw response object, false for the data from the response
+ * @returns Raw response or data from response based on rawResponse argument from fetch call
+ */
+// eslint-disable-next-line import/prefer-default-export
+export const runFetchHelper = async <Params extends BaseUriParameters>(
```
bendvc, line 29: I'd opts for a more commonly used name for a fetch helper here, aka "doFetch" unless there is some specific precedence that we end all our helpers with "Helper". Also "run" is typically used for scripts, services.

templates/operations.ts.hbs
```diff
@@ -187,26 +187,24 @@
         headers[USER_AGENT_HEADER] = [headers[USER_AGENT_HEADER], USER_AGENT_VALUE].join(" ");
       }
 
-      const requestOptions = {
-        ...this.clientConfig.fetchOptions,
-        {{#if (isRequestWithPayload request)}}body: this.clientConfig.transformRequest(options.body, headers),{{/if}}
-        headers,
-        method: "{{loud method}}"
-      };
+      const response = await runFetchHelper(
```
bendvc, line 190: Were their any measurable change in the side of a generated lib before and after with this change. Looks like we are only saving 2 lines roughly per use of this template, but it might add up.

src/static/helpers/fetchHelper.test.ts
```diff
@@ -0,0 +1,116 @@
...
+ * SPDX-License-Identifier: BSD-3-Clause
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+import nock from 'nock';
+import {Response} from 'node-fetch';
+import * as environment from './environment';
+import ClientConfig from '../clientConfig';
+import {runFetchHelper} from './fetchHelper';
+
+describe('runFetchHelper', () => {
```
bendvc, line 13: The fetch helper accepts a client config and uses it to assign headers amongst other things right? Do we want to test the scenario where we are overwriting a header value, which valve takes precedence? Lets lock it down with a test

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # CHANGELOG
 
+## v1.14.0-dev
+
+#### Enchancements
+
+- Add support for custom APIs [#149](https://github.com/SalesforceCommerceCloud/commerce-sdk-isomorphic/pull/149)
```
bendvc, line 7: Just being nit picky here. But it might be preferential to be specific here, that we added a helper to use to calling custom api's. Otherwise people might assuming we did more and wonder why the client doesn't have specific API classes for their environments custom api's.

templates/operations.ts.hbs
joeluong-sfcc: `yarn run check:size`  for `main` branch:
```
 PASS  lib/index.cjs.js: 44.84KB < maxSize 45KB (gzip) 

 PASS  lib/index.esm.js: 44.73KB < maxSize 45KB (gzip) 

 PASS  commerce-sdk-isomorphic-with-deps.tgz: 337.06KB < maxSize 350KB (gzip)
```

`yarn run check:size`  for `ju/custom-endpoint` branch:
```
PASS  lib/index.cjs.js: 44.13KB < maxSize 45KB (gzip) 

PASS  lib/index.esm.js: 44.02KB < maxSize 45KB (gzip) 

PASS  commerce-sdk-isomorphic-with-deps.tgz: 337.43KB < maxSize 350KB (gzip) 
```

Seems like `lib/index.cjs.js` and `lib/index.esm.js` have decreased slightly, but `commerce-sdk-isomorphic-with-deps.tgz` has increased slightly 😮

src/static/helpers/fetchHelper.ts
```diff
@@ -0,0 +1,68 @@
...
+): Promise<Response | unknown> => {
+  const headers: Record<string, string> = {
+    ...clientConfig?.headers,
+    ...options?.headers,
+  };
+
+  const requestOptions: FetchOptions = {
+    ...clientConfig?.fetchOptions,
+    headers,
+    body: options?.body as BodyInit & string,
```
unandyala, line 49: can you clarify why we need intersection type here?

joeluong-sfcc: this typing is actually incorrect as I misunderstood the body type with node/browser fetch, but I've updated it and added a clarifying comment

unandyala (approved):

unandyala (approved):

README.md
```diff
@@ -141,6 +141,62 @@ const searchResult = await shopperSearch.productSearch({
...
+    apiName: 'loyalty-info',
+    apiVersion: 'v1', // defaults to v1 if not provided
+  },
+};
+
+// Flag to retrieve raw response or data from helper function
+const rawResponse = false;
+const accessToken = '<INSERT ACCESS TOKEN HERE>';
+
+let response = await helpers.callCustomEndpoint(
```
johnboxall, line 177: I don't feel strongly, but would you consider an alternative argument structure for `callCustomEndpoint`?

```js
await helpers.callCustomEndpoint({
  options: myOptions,
  clientConfig: myClientConfig,
  rawResponse: true
})
```

This interface makes it much more clear to folks using regular old JavaScript what the impact of the three arguments will be.

It would also allow you to significantly shrink this code example:

```js
await helpers.callCustomEndpoint({
  options: {
    method: "GET",
    parameters: {
      queryParameter: "queryParameter1",
      siteId: SITE_ID,
    },
    headers: {
      "Content-Type": "application/json",
      authorization: `Bearer <your-access-token>`,
    },
  },
  clientConfig: {
    clientId: "<your-client-id>",
    organizationId: "<your-org-id>",
    shortCode: "<your-short-code>",
    siteId: "<your-site-id>",
    // Custom API path parameters
    endpointPath: "customers",
    apiName: "loyalty-info",
    apiVersion: "v1", // defaults to v1 if not provided
  },
  rawResponse: true,
});
```

README.md
```diff
@@ -141,6 +141,62 @@ const searchResult = await shopperSearch.productSearch({
...
+  {
+    method: 'GET',
+    parameters: {
+      queryParameter: 'queryParameter1',
+      siteId: SITE_ID,
+    },
+    headers: {
+      'Content-Type': 'application/json',
```
johnboxall, line 185: Do we expect callers to pass the content type or can it be automatically inferred or default to JSON?

README.md
```diff
@@ -141,6 +141,62 @@ const searchResult = await shopperSearch.productSearch({
...
+  clientConfig,
+  rawResponse
+)
+
+console.log('RESPONSE: ', response)
+```
+
+For more documentation about this helper function, please refer to the [commerce-sdk-isomorphic docs](https://salesforcecommercecloud.github.io/commerce-sdk-isomorphic/modules/helpers.html).
+
+For more information about custom APIs, please refer to the [Salesforce Developer Docs](https://developer.salesforce.com/docs/commerce/commerce-api/guide/custom-apis.html?q=custom+API)
```
johnboxall, line 198: Again, remove the search from this link.

README.md
johnboxall: Would you also consider showing an example of a mutation request eg. a POST where you pass a request body?

src/static/helpers/customApi.ts
```diff
@@ -0,0 +1,94 @@
...
+ * @param options.body? - Body that is used for the request
+ * @param clientConfig - Client Configuration object used by the SDK with properties that can affect the fetch call
+ * @param clientConfig.parameters - Path parameters used for custom API endpoints. The required properties are: apiName, endpointPath, organizationId, and shortCode. An error will be thrown if these are not provided.
+ * @param clientConfig.headers? - Additional headers that are added to the request. Authorization header should be in this argument or in the options?.headers. options?.headers will override any duplicate properties.
+ * @param clientConfig.baseUri? - baseUri used for the request, where the path parameters are wrapped in curly braces. Default value is 'https://{shortCode}.api.commercecloud.salesforce.com/custom/{apiName}/{apiVersion}'
+ * @param clientConfig.fetchOptions? - fetchOptions that are passed onto the fetch request
+ * @param clientConfig.throwOnBadResponse? - flag that when set true will throw a response error if the fetch request fails
```
johnboxall, line 35: What does "the fetch request fails" mean specifically? It would likely be useful to say!

README.md
```diff
@@ -141,6 +141,62 @@ const searchResult = await shopperSearch.productSearch({
...
+const clientConfig = {
+  parameters: {
+    clientId: CLIENT_ID,
+    organizationId: ORG_ID,
+    shortCode: SHORT_CODE,
+    siteId: SITE_ID,
+    // Custom API path parameters
+    endpointPath: 'customers',
+    apiName: 'loyalty-info',
+    apiVersion: 'v1', // defaults to v1 if not provided
```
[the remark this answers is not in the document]
joeluong-sfcc: Added a new property to `options` called `customApiPathParameters`

```
customApiPathParameters?: {
    apiName?: string;
    apiVersion?: string;
    endpointPath?: string;
    organizationId?: string;
    shortCode?: string;
  };
```

path parameters can now be passed either in `options.customApiPathParameters` or `clientConfig.parameters`.

README.md
joeluong-sfcc: Updated implementation to default to `"Content-Type" : "application/json"` if `Content-Type` header is not provided

README.md
```diff
@@ -141,6 +141,88 @@ const searchResult = await shopperSearch.productSearch({
...
+};
+
+// Required params: apiName, endpointPath, shortCode, organizaitonId
+// Required path params can be passed into:
+// options.customApiPathParameters or clientConfig.parameters
+const customApiArgs = { 
+  endpointPath: 'customers',
+  apiName: 'loyalty-info',
+  apiVersion: 'v1', // defaults to v1 if not provided
+}
```
johnboxall, line 176: Seek to order relative to the order they appear in the URL:

```
https://{shortCode}.api.commercecloud.salesforce.com/custom/{apiName}/{apiVersion}/organizations/{organizationId}/{endpointPath
```

```suggestion
const customApiArgs = { 
  apiName: 'loyalty-info',
  apiVersion: 'v1', // defaults to v1 if not provided
  endpointPath: 'customers',
}
```

README.md
```diff
@@ -141,6 +141,88 @@ const searchResult = await shopperSearch.productSearch({
...
+
+// client configuration parameters
+const clientConfigExample = {
+  parameters: {
+    clientId: CLIENT_ID,
+    organizationId: ORG_ID,
+    shortCode: SHORT_CODE,
+    siteId: SITE_ID,
+  }
+};
```
johnboxall, line 167: I don't feel strongly, but you're repeating yourself here:

```suggestion
const clientConfigExample = {
  parameters: {
    clientId: <your-client-id>,
    organizationId: "<your-org-id>",
    shortCode: "<your-short-code>",
    siteId: "<your-site-id>",
  }
};
```

Also, because you've chosen meaningful variable names, you don't really need the code comment.

Why do you pass `siteId` as a param to the client Config and then pass it again as a `parameter` later?

johnboxall (commented):
@joeluong-sfcc – the updated interface looks good. I've only reviewed the README, I'll let the rest up to others.

README.md
joeluong-sfcc: That was a miss where I didn't use the siteId from the clientConfig, code has been updated to use siteId as query parameter if available

src/static/helpers/customApi.ts
```diff
@@ -0,0 +1,150 @@
...
+};
+
+export interface CustomParams {
+  apiName?: string;
+  apiVersion?: string;
+  endpointPath?: string;
+  organizationId?: string;
+  shortCode: string;
+  [key: string]: unknown;
+}
```
bendvc, line 34: Is it possible to extend from any existing types? I'm assuming there is a type that includes the org id and short code, etc. If that type has to change, we would have to change it in 2 places now.

bendvc: Additionally, why are the params optional?

src/static/helpers/customApi.ts
```diff
@@ -0,0 +1,150 @@
...
+        `Missing required property needed in options.customApiPathParameters or clientConfig.parameters: ${arg}`
+      );
+    }
+  });
+
+  if (!pathParams.apiVersion) {
+    pathParams.apiVersion = 'v1';
+  }
+
+  const defaultBaseUri =
```
bendvc, line 103: We had the conversation about the baseUri before. It doesn't look like we are going to be do anything about handling non-prod uri's. Is that the case, and have you shared that decision with anyone else or documented it.

bendvc, line 103: We previously talked about this decision on hard coding prod uri.. are we cool with that. Have we documented it anywhere?

joeluong-sfcc: It's the default base URI if you don't provide one yourself. You can override this behavior by passing in your own baseUri in the clientConfig

```
const clientConfig = {
  parameters: {
      ...
  },
  baseUri: 'https://{shortCode}.alternativeBaseUri.com/custom/{apiName}/{apiVersion}'
};
```

Behavior is documented in the tsdoc comments. When we release and run the generate doc command it'll populate the static docs site:

```
 * @param args.clientConfig.baseUri? - baseUri used for the request, where the path parameters are wrapped in curly braces. Default value is 'https://{shortCode}.api.commercecloud.salesforce.com/custom/{apiName}/{apiVersion}'

```

src/static/helpers/customApi.ts
joeluong-sfcc: The only existing interface is `BaseUriParameters`:

```
export interface BaseUriParameters {
  shortCode: string;
  version?: string; // Optional, will default to "v1" if not provided.
}
```

and we're required to extend it when passing in a type for client config:

```
/**
 * Base options that can be passed to the `ClientConfig` class.
 */
export interface ClientConfigInit<Params extends BaseUriParameters> {
  ...
  parameters: Params;
}
```

joeluong-sfcc: Params are optional as they can be passed in through `clientConfig.parameters` or `options.parameters`. We'll throw an error if required parameters are missing

templates/operations.ts.hbs
```diff
@@ -187,26 +187,24 @@
...
       }
-      {{#unless (eq (getReturnTypeFromOperation this) "void") }}
-        else {
-          const text = await response.text();
-          // It's ideal to get "{}" for an empty response body, but we won't throw if it's truly empty
-          return text ? JSON.parse(text) : {};
-        }
-      {{/unless}}
+      {{else}}
+      return response as Response | {{getReturnTypeFromOperation this}};
```
bendvc, line 206: Can you explain this line a little bit more for me. If we are expecting that we aren't using a _rawResponse_ we are saying that the return value is the **union** of a Response object and the return type of this operation? Why wouldn't it simply be the return type of the operation without the union?

src/static/helpers/customApi.ts
bendvc: I guess my question is, and I think I know the answer. But if you were to generate a client where the raml specifies the base url a something like 'https://{shortCode}.dev-api.commercecloud.salesforce.com/custom/{apiName}/{apiVersion}' notice the "dev" in "dev-api" all calls to the apis will goto that url. but if I make a custom api call without passing that "dev" value, it will make it to production. Correct?

joeluong-sfcc: Yes, because the `callCustomEndpoint` helper function does not interact with any type of RAML, it'll use the production URI as default if you don't provide a base URI.

templates/operations.ts.hbs
joeluong-sfcc: We use the `rawResponse` when we call the `doFetch` helper function for the `response` variable. Since we don't know what `rawResponse` is when we pass it to `doFetch`, the result is either a response object OR the return type

unandyala: I think it is to handle any unexpected response - the response that is not defined in RAML specially in case of error response

src/static/helpers/customApi.ts
unandyala: May be we shouldn't have default URI. How about just throwing an error if the baseUri is not provided?

joeluong-sfcc: Moved the default base URI into a config file. I think it still makes sense to have a default URI as all the other APIs use the production URI as default, but I've made it more clear in the README example on how to pass it into the function. Does that work for y'all @unandyala @bendvc?

unandyala (approved):
LGTM

bendvc (approved):

PR #150, opened 2024-04-29
@W-15248536@ - [v3] Google Search Console fix createCodeVerifier

adamraya:
**Problem**
When Google crawler calls[ createCodeVerifier()](https://github.com/SalesforceCommerceCloud/pwa-kit/blob/37c932ec02d83fe7961ed1b9cb8717eea9952434/packages/template-retail-react-app/app/commerce-api/pkce.js#L22) to create the code verifier we use for the PKCE authentication flow, the result from nanoid() is the same in two different runs.  This causes the same generated code_challenge to be used twice thus Google crawler gets the error stating that the PKCE code_challenge must be unique error.

The root of this issue is the fact that Googlebot’s implementation of random() is deterministic, meaning that it won’t produce unique random values from one string of runs to the next.

The result is that Google crawler indexes the error page instead of the proper page.

**Solution**
Add entropy to nanoid() using the library `seedrandom` to avoid duplicated code_challenge sent to SCAPI by Google crawler browser. 

Nanoid docs on [Custom Random Bytes Generator](https://github.com/ai/nanoid?tab=readme-ov-file#custom-random-bytes-generator).

**How to Test-Drive  This PR**

1. Follow the steps to reproduce in the quip doc: https://salesforce.quip.com/U1k5A2fXzpym#temp:C:DXJd70645fbe1a44e999dff07b74
2. Verify Google Search Console can successfully index the indicated URLs.

### DIFF
--- a/package.json
+++ b/package.json
@@ -97,7 +97,8 @@
   },
   "dependencies": {
     "nanoid": "^3.3.4",
-    "node-fetch": "2.6.12"
+    "node-fetch": "2.6.12",
+    "seedrandom": "^3.0.5"
   },
   "devDependencies": {
     "@babel/cli": "7.18.6",
@@ -117,6 +118,7 @@
     "@types/handlebars-helpers": "^0.5.3",
     "@types/node-fetch": "^2.6.2",
     "@types/react-dom": "^16.9.16",
+    "@types/seedrandom": "^3.0.8",
     "@typescript-eslint/eslint-plugin": "^4.33.0",
     "@typescript-eslint/parser": "^4.33.0",
     "autoprefixer": "9.8.8",
@@ -179,7 +181,7 @@
     },
     {
       "path": "commerce-sdk-isomorphic-with-deps.tgz",
-      "maxSize": "350 kB"
+      "maxSize": "400 kB"
     }
   ],
   "proxy": "https://SHORTCODE.api.commercecloud.salesforce.com"
--- a/package.json
+++ b/package.json
@@ -177,7 +177,7 @@
   "bundlesize": [
     {
       "path": "lib/**/*.js",
-      "maxSize": "45 kB"
+      "maxSize": "46 kB"
     },
     {
       "path": "commerce-sdk-isomorphic-with-deps.tgz",
--- a/src/static/helpers/slasHelper.ts
+++ b/src/static/helpers/slasHelper.ts
@@ -5,8 +5,8 @@
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
 
-import {nanoid} from 'nanoid';
-
+import {customRandom, urlAlphabet} from 'nanoid';
+import seedrandom, {PRNG} from 'seedrandom';
 import {isBrowser} from './environment';
 
 import {
@@ -39,11 +39,18 @@ export const getCodeAndUsidFromUrl = (
   };
 };
 
+const nanoid = (): string => {
+  const rng: PRNG = seedrandom(String(+new Date()), {entropy: true});
+  return customRandom(urlAlphabet, 128, size =>
+    new Uint8Array(size).map(() => 256 * rng())
+  )();
+};
+
 /**
  * Creates a random string to use as a code verifier. This code is created by the client and sent with both the authorization request (as a code challenge) and the token request.
  * @returns code verifier
  */
-export const createCodeVerifier = (): string => nanoid(128);
+export const createCodeVerifier = (): string => nanoid();
 
 /**
  * Encodes a code verifier to a code challenge to send to the authorization endpoint
--- a/src/static/helpers/slasHelper.ts
+++ b/src/static/helpers/slasHelper.ts
@@ -39,6 +39,10 @@ export const getCodeAndUsidFromUrl = (
   };
 };
 
+/**
+ * Adds entropy to nanoid() using seedrandom to ensure that the code_challenge sent to SCAPI by Google's crawler browser is unique.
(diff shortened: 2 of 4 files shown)


src/static/helpers/slasHelper.ts
```diff
@@ -39,11 +39,18 @@ export const getCodeAndUsidFromUrl = (
   };
 };
 
+const nanoid = (): string => {
+  const rng: PRNG = seedrandom(String(+new Date()), {entropy: true});
+  return customRandom(urlAlphabet, 128, size =>
+    new Uint8Array(size).map(() => 256 * rng())
+  )();
+};
```
johnboxall, line 47: Consider a comment comment on why this random seeding is nessecary given that browsers should handle it.

joeluong-sfcc:
`yarn run check:size` results:

current `main` branch:

```
 PASS  lib/index.cjs.js: 44.84KB < maxSize 45KB (gzip) 

 PASS  lib/index.esm.js: 44.73KB < maxSize 45KB (gzip) 

 PASS  commerce-sdk-isomorphic-with-deps.tgz: 338.08KB < maxSize 350KB (gzip) 
```

`google-search-console-fix-v3` branch:
```
 PASS  lib/index.cjs.js: 44.96KB < maxSize 46KB (gzip) 

 PASS  lib/index.esm.js: 44.84KB < maxSize 46KB (gzip) 

 PASS  commerce-sdk-isomorphic-with-deps.tgz: 390.52KB < maxSize 400KB (gzip)
```

We go from `338.08KB` to `390.52KB` which is a decent jump for `commerce-sdk-isomorphic-with-deps.tgz`. We should try to make this package should be as lightweight as possible so just wanted to ask a few things:

1. Might be a silly question, but is it possible for us to add entropy without the use of a third party package?
2. Is there a specific reason why we use `seedrandom` versus an alternative package that provides similar functionality? Mostly asking if there's an alternative package we could use that is smaller in size

This PR looks good otherwise, just wanted to check for the reasoning behind the implementation

joeluong-sfcc (commented):
`yarn run check:size` results:

current `main` branch:

```
 PASS  lib/index.cjs.js: 44.84KB < maxSize 45KB (gzip) 

 PASS  lib/index.esm.js: 44.73KB < maxSize 45KB (gzip) 

 PASS  commerce-sdk-isomorphic-with-deps.tgz: 338.08KB < maxSize 350KB (gzip) 
```
`google-search-console-fix-v3` branch:

```
 PASS  lib/index.cjs.js: 44.96KB < maxSize 46KB (gzip) 

 PASS  lib/index.esm.js: 44.84KB < maxSize 46KB (gzip) 

 PASS  commerce-sdk-isomorphic-with-deps.tgz: 390.52KB < maxSize 400KB (gzip)
 ```
We go from 338.08KB to 390.52KB which is a decent jump for `commerce-sdk-isomorphic-with-deps.tgz`. We should try to make this package should be as lightweight as possible so just wanted to ask a few things:

1. Might be a silly question, but is it possible for us to add entropy without the use of a third party package?
2. Is there a specific reason why we use `seedrandom` versus an alternative package that provides similar functionality? Mostly asking if there's an alternative package we could use that is smaller in size

This PR looks good otherwise, just wanted to check for the reasoning behind the implementation

**Comment by adamraya** [comment]
> We go from 338.08KB to 390.52KB which is a decent jump for `commerce-sdk-isomorphic-with-deps.tgz`. We should try to make this package should be as lightweight as possible so just wanted to ask a few things:
> 
> 1. Might be a silly question, but is it possible for us to add entropy without the use of a third party package?
> 2. Is there a specific reason why we use `seedrandom` versus an alternative package that provides similar functionality? Mostly asking if there's an alternative package we could use that is smaller in size
> 
> This PR looks good otherwise, just wanted to check for the reasoning behind the implementation

No silly question at all. I did notice the same. I think it's a good convo to have before making a decision.

The reasons to use `seedrandom`.
1. It's documented on the nanoid docs. https://github.com/ai/nanoid?tab=readme-ov-file#custom-random-bytes-generator
2. It's a popular NPM package with a high number of downloads. https://www.npmjs.com/package/seedrandom


We need to find an alternative that doesn't rely on Math.random() to add entropy which is the source of the issue. I didn't find an alternative to seedrandom.

The Google Craweler browser doesn't use a deterministic value when using seedrandom. That's why we get a new number from one run to the other.

**Review by joeluong-sfcc — DISMISSED** [review]
Thought about this some more, I think the size change should be ok since the minified bundle size is the size that we care more about, and that has only increased slightly.

File: package.json [hunk 2]
```diff
@@ -175,11 +177,11 @@
   "bundlesize": [
     {
       "path": "lib/**/*.js",
-      "maxSize": "45 kB"
+      "maxSize": "46 kB"
     },
     {
       "path": "commerce-sdk-isomorphic-with-deps.tgz",
-      "maxSize": "350 kB"
+      "maxSize": "400 kB"
```
bendvc, line 184: Is this because that lib has a lot of docs or large source that is being added to the gzip?

adamraya: I assume that's the reason as you can see in the previous [Joel's comment](https://github.com/SalesforceCommerceCloud/commerce-sdk-isomorphic/pull/150#pullrequestreview-2034095413) comparing the before and after file sizes. We added two libraries `seedrandom` as a dependency and `@types/seedrandom` as a devDependency.

But as later [Joel also mentions](https://github.com/SalesforceCommerceCloud/commerce-sdk-isomorphic/pull/150#pullrequestreview-2036226789) the minimized bundle size is the size we care about and that one only increased slightly, 1KB.

joeluong-sfcc (approved):

src/static/helpers/slasHelper.ts
adamraya: Done.

bendvc (approved):

PR #153, opened 2024-05-09
Release v2.0.0 - @W-15675100@

joeluong-sfcc:
This PR Releases `v1.14.0` of the `commerce-sdk-isomorphic`. Changes include:

- Bug fix for createCodeVerifier
- Custom API support

This PR pulls the latest RAML to date from anypoint exchange by modifying the local version of raml-toolkit: https://github.com/SalesforceCommerceCloud/raml-toolkit/pull/223/files. As a result, the changes are quite significant with this PR as we're pulling a lot of data from being out of data with our APIs. 

API Changes:
- ~Shopper Baskets~ Excluding this change as we need to support shopper baskets v2 first
    - ~2.0.3 --> 2.0.7~
- Shopper Context
    - 0.0.23 --> 0.0.26
    - Exchange module updates
    - optional clientIp arg added 
- Shopper Customers
    - 0.0.43 --> 0.0.48
    - Removal of deprecated endpoints
        - invalidateCustomerAuth
        - authorizeCustomer
        - authorizeTrustedSystem
    - Removal of headers
- Shopper gift certificates
    - 1.0.12 --> 1.0.15
    - Exchange module updates
    - Headers were removed
- Shopper orders
    - 0.3.5 --> 0.4.7
    - Operation guestOrderLookup added
    - Headers removed
    - Exchange module updates
- Shopper Products
    - 0.0.25 --> 0.0.32
    - Select enum options updated
    - Category Type updated with optional property
    - Headers removed
    - Exchanged module updates
- Shopper Promotions
    - 1.0.21 --> 1.0.25
    - Exchange modules updated
    - Headers removed
    - Response type updated
        - **count renamed to limit**
        - **this may be breaking** - UPDATE: This change has been released 2 years ago, and is how the API currently is, so I think we'll be ok releasing this
- Shopper Search
    - 1.0.38 --> 1.0.40
    - Exchange modules updated
- Shopper SEO
    - 1.0.6 --> 1.0.8
    - Exchange modules updated
- Shopper Stores
    - 1.0.13
    - API added


apis/shopper-promotions/shopper-promotions.raml
```diff
@@ -68,7 +68,7 @@ types:
     examples:
       promotion: !include Examples/promotions_result.raml
     properties:
-      count:
+      limit:
```
kevinxh, line 71: it seems `count` is renamed to `limit`? is this expected? isn't this a breaking change?

kevinxh:
Your PR description mentioned headers are removed, which headers are removed?sorry it's hard to tell from the diff

apis/shopper-promotions/shopper-promotions.raml
joeluong-sfcc: This change was made around 2 years ago where it was updated from `count` to `limit`. Technically it is a breaking change, but the API doesn't return `count` but instead returns `limit` and has been for 2 years. It might be more accurate to categorize this as a bug fix

joeluong-sfcc:
> Your PR description mentioned headers are removed, which headers are removed?sorry it's hard to tell from the diff

@kevinxh 

Headers such as:
- `X-RateLimit-Limit`
- `X-RateLimit-Remaining`
- `X-RateLimit-Reset`
- `Retry-After`

Since the SDKs don't consider headers when generating the typescript files, I think we should be good to ignore those changes

src/test/crossFetchNode.test.ts
```diff
@@ -19,107 +19,10 @@ type TestConfigParameters = typeof config.parameters;
  * Both POST and GET are exercised to ensure compatibility.
  */
 
-const customerClient = new ShopperCustomers(config);
 const searchClient = new ShopperSearch(config);
 
 beforeEach(() => nock.cleanAll());
 
-test('test getting a token with a post operation', async () => {
```
joeluong-sfcc, line 27: This is the old deprecated way to retrieve access tokens. Since we have tests for the slas helpers, I think we can remove these tests that call `authorizeCustomer`

kevinxh (dismissed):

vcua-mobify:
With all of the API changes (some of which seems like they are breaking changes), should this be a major version release?

joeluong-sfcc:
> With all of the API changes (some of which seems like they are breaking changes), should this be a major version release?

I think we're ok with a minor release since the 'breaking changes' are:
- Shopper Products - count is renamed to limit
    - This is more of a bug fix since the API has been like this for 2 years, we just never updated on our side
- Removal of certain `Shopper Customers` deprecated endpoints
    - These endpoints were already deprecated and no customers should be using these anyways, everyone has migrated to SLAS

but I could be open to doing a major version bump, but I'd like to take some time to consider any other breaking changes we'd like to include if that's the case

alexvuong (approved):

vmarta (approved):

src/test/parameters.test.ts
```diff
@@ -6,7 +6,7 @@
  */
 
 import nock from 'nock';
-import {ShopperCustomers} from '../lib/shopperCustomers';
+import {ShopperSearch} from '../lib/shopperSearch';
```
unandyala, line 9: Did we do this change as the actions/login is deprecated?

templates/client.ts.hbs
```diff
@@ -1,7 +1,7 @@
 import ClientConfig, { ClientConfigInit } from "./clientConfig";
 // Must not import from ./helpers/index to avoid circular dependency via ShopperLogin
 import { isBrowser } from "./helpers/environment";
-import { doFetch } from "./helpers";
+import { doFetch } from "./helpers/fetchHelper";
```
unandyala, line 4: why do we need this changes?

vmarta (approved):

src/test/parameters.test.ts
joeluong-sfcc: Yes that is correct

templates/client.ts.hbs
joeluong-sfcc: `// Must not import from ./helpers/index to avoid circular dependency via ShopperLogin`

We import directly from the file to prevent any circular dependency issues

Issue #108, opened 2023-01-12
Does not run in react-native


anutting:
I installed this module in my react-native project, but receive an error:

```
error: Error: While trying to resolve module `crypto` from file `my-project/node_modules/commerce-sdk-isomorphic/lib/index.cjs.js`, 
```

It seems like this module requires node's crypto package that is not available in react-native.

Please provide a workaround, alternative module, or fix for this issue I can use the commerce-sdk in react-native.

Thanks you

joeluong-sfcc:
Hi @anutting, thanks for creating this issue. A work item in our backlog has been created and we'll loop you back in with any updates.

asad-pwc:
Hi,
Is there any update on this issue, since I am working on react-native and still facing this issue

bendvc:
> Hi, Is there any update on this issue, since I am working on react-native and still facing this issue

Hi @asad-pwc, I don't think the team will have time to look into this as we do not normally use react native and would have to ramp up. However it does look like this problem isn't specific to our lib, there is a shim that was created [here](https://www.npmjs.com/package/react-native-crypto) that you can probably use to resolve your problem. Have you given that a try?

FerVillanuevas:
i already fix this.

The issue its on helpers, because RN did not hava a Crypto lib, insted of expo-crypto.

So we should create an alieas `expo-crypto` to `crypto`

To do that install this plugin https://www.npmjs.com/package/babel-plugin-module-resolver
and update the configuration on: babel.config.js
`plugins: [
      [
        "module-resolver",
        {
          alias: {
            crypto: "expo-crypto",
          },
        },
      ],
    ],`

Then on the library we should verify if is RN, Brower or Node.

`if (
      typeof navigator !== 'undefined' &&
      navigator.product === 'ReactNative'
    ) {
      challenge =
        (await crypto.digestStringAsync?.('SHA-256', codeVerifier, {
          encoding: 'base64',
        })) || '';
    } else {
      challenge = urlSafe(
        crypto.default
          .createHash('sha256')
          .update(codeVerifier)
          .digest('base64')
      );
    }`

i will create a PR, but dont know if will get aprroved

FerVillanuevas:
https://github.com/SalesforceCommerceCloud/commerce-sdk-isomorphic/pull/154 PR here

samskinner-joann:
[...]

PR #154, opened 2024-05-17
Expo-Crypto for RN

FerVillanuevas:
For RN there is not a Crypto library.

So we should use expo-crypto and make an alias to crypto

## RN Config

# Install [Bable Module Resolver](https://www.npmjs.com/package/babel-plugin-module-resolver)
# Update babel.config.js
`module.exports = function (api) {
  api.cache(true);
  return {
    presets: ["babel-preset-expo"],
    plugins: [
      [
        "module-resolver",
        {
          alias: {
            crypto: "expo-crypto",
          },
        },
      ],
    ],
  };
};
`

So we are telling to RN that the crypto lib is expo-crypto insted of crypto, the fact is expo-crypto has different methods, so we need to verify if is RN or Brower/Node then if is react native call the methods from expo.

There is an issue, navigator.product its depreacted, but other that this config works as expect and all test passed

### DIFF
--- a/src/static/helpers/slasHelper.ts
+++ b/src/static/helpers/slasHelper.ts
@@ -93,10 +93,15 @@ export const generateCodeChallenge = async (
       typeof navigator !== 'undefined' &&
       navigator.product === 'ReactNative'
     ) {
-      challenge =
-        (await crypto.digestStringAsync?.('SHA-256', codeVerifier, {
+      const base64Digest = await crypto.digestStringAsync?.(
+        'SHA-256',
+        codeVerifier,
+        {
           encoding: 'base64',
-        })) || '';
+        }
+      );
+
+      challenge = base64Digest ? urlSafe(base64Digest) : '';
     } else {
       challenge = urlSafe(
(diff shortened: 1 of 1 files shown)


FerVillanuevas:
Fix crypto on RN

https://github.com/SalesforceCommerceCloud/commerce-sdk-isomorphic/issues/108

joeluong-sfcc:
Hi @FerVillanuevas, just to clarify, adding this fix does not work with react native out of the box, is that correct? Additional configuration is needed on the react native side on top of this fix in order for the `commerce-sdk-isomorphic` to work with react native. 

Ideally things would work out of the box for all environments, browser/node/react native without additional configuration. In your manual testing, were you able to find a library that provides the same functionality as `crypto` that works on all environments, potentially something like `crypto-js`?

FerVillanuevas:
@joeluong-sfcc this won’t fix anything I will close it,

The real fix it’s move helpers outside of isometric, ideal on sdk-react, that why crypto it’s not part of this, works on all platforms

PR #157, opened 2024-06-04
Export parameter keys (@W-15676813@)

vmarta:
This PR exports the parameter keys for each API method, so that a client like `commerce-sdk-react` will be able to rely on them (and no longer hardcoding those param keys).

For an example on how to consume the exported parameter keys, please see https://github.com/SalesforceCommerceCloud/pwa-kit/pull/1812

Note: this PR is for the parameter _keys_ (not its Typescript types). So we're exporting the list/array.

## How to test-drive
In this case, I worked with 2 separate repos: this commerce-sdk-isomorphic (exports the parameter keys) and the pwa-kit (to consume those keys). 

So first of all, let's build the commerce-sdk-isomorphic to get its minified output:
1. `yarn install`
2. `yarn run renderTemplates`
3. `yarn run build:lib`
4. If you need more details: https://github.com/SalesforceCommerceCloud/commerce-sdk-isomorphic/blob/export-parameter-keys/Contributing.md

Once we've gotten those minified scripts, we need to somehow use them in the pwa-kit repo. I've tried `npm link` but gotten an error. I'm not sure how to `npm link` properly within a monorepo like ours. Anyways, I manually created a symlink instead, like this:
1. Switch to the pwa-kit repo and check out this sample PR: https://github.com/SalesforceCommerceCloud/pwa-kit/pull/1812
2. Create symlink from commerce-sdk-react
```bash
	# From "pwa-kit/packages/commerce-sdk-react"
	rm -r node_modules/commerce-sdk-isomorphic/lib
	ln -s ~/repos/commerce-sdk-isomorphic/lib node_modules/commerce-sdk-isomorphic
```
3. Now in your code editor, open the file "packages/commerce-sdk-react/src/hooks/ShopperProducts/query.ts".
4. Inspect the values of some variables, and you'll see the parameter keys like this:

iTerm2 2024-06-04 at 11 04 02

iTerm2 2024-06-04 at 11 04 43

### DIFF
--- a/package.json
+++ b/package.json
@@ -177,11 +177,11 @@
   "bundlesize": [
     {
       "path": "lib/**/*.js",
-      "maxSize": "46 kB"
+      "maxSize": "48 kB"
     },
     {
       "path": "commerce-sdk-isomorphic-with-deps.tgz",
-      "maxSize": "400 kB"
+      "maxSize": "415 kB"
     }
   ],
   "proxy": "https://SHORTCODE.api.commercecloud.salesforce.com"
--- a/templates/client.ts.hbs
+++ b/templates/client.ts.hbs
@@ -110,3 +110,5 @@ const throwMissingRequiredParamError = ({name, paramType}: {name: string, paramT
 const warnInvalidParam = ({paramName, methodName}: {paramName: string, methodName: string}) => {
   console.warn(`Invalid Parameter for ${methodName}: ${paramName}`)
 }
+
+{{> paramKeysPartial model}}
--- a/templates/client.ts.hbs
+++ b/templates/client.ts.hbs
@@ -80,12 +80,12 @@ export type {{name.upperCamelCase}}Parameters = {{name.upperCamelCase}}PathParam
 *
 * <span style="font-size:.7em; display:block; text-align: right">
 * API Version: {{metadata.version}}<br />
-* Last Updated: {{metadata.updatedDate}}<br />  
+* Last Updated: {{metadata.updatedDate}}<br />
 * </span>
 {{#if (eq (lowercase metadata.categories.[CC Version Status].[0]) "beta")}}
-* @beta 
+* @beta
 * {{/if}}
-* 
+*
 
 */
 export class {{name.upperCamelCase}}<ConfigParameters extends {{{name.upperCamelCase}}}Parameters & Record<string, unknown>> {
@@ -102,5 +102,11 @@ export class {{name.upperCamelCase}}<ConfigParameters extends {{{name.upperCamel
   }
 
   {{> operationsPartial model}}
+}
 
+const throwMissingRequiredParamError = ({name, paramType}: {name: string, paramType: 'query' | 'path'}) => {
+  throw new Error(`Missing required ${paramType} parameter: ${name}`);
+}
+const warnInvalidParam = ({paramName, methodName}: {paramName: string, methodName: string}) => {
+  console.warn(`Invalid Parameter for ${methodName}: ${paramName}`)
 }
--- a/templates/index.ts.hbs
+++ b/templates/index.ts.hbs
@@ -1,5 +1,5 @@
 {{#each children}}
-export { {{name.upperCamelCase}} } from "./{{name.lowerCamelCase}}";
+export { {{name.upperCamelCase}}, paramKeys as {{name.upperCamelCase}}ParamKeys } from "./{{name.lowerCamelCase}}";
 import type * as {{name.upperCamelCase}}Types from "./{{name.lowerCamelCase}}";
 export type { {{name.upperCamelCase}}Types };
 {{/each}}
--- a/templates/operations.ts.hbs
+++ b/templates/operations.ts.hbs
@@ -1,5 +1,6 @@
 {{#each encodes.endPoints}}
   {{#each operations}}
+
     static readonly {{{name}}}ParamKeys = [
       {{#each ../parameters}}
       '{{{name}}}',
--- a/templates/operations.ts.hbs
+++ b/templates/operations.ts.hbs
@@ -155,7 +155,7 @@
         shortCode: configParams.shortCode,
         version: configParams.version || "v1"
       };
-      // TODO
+      // TODO: try extracting this
       {{{@root.name.upperCamelCase}}}.{{{name}}}PathParamKeys.forEach((paramKey) => {
         const value = optionParams[paramKey] ?? configParams[paramKey];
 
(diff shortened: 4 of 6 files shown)


templates/operations.ts.hbs
```diff
@@ -1,5 +1,26 @@
...
+    static readonly {{{name}}}ParamKeysRequired = [
+      {{#each ../parameters}}
+      '{{{name}}}',
+      {{/each}}
+      {{#each request.queryParameters}}
+      {{#if (is required "true")}}
+      '{{{name}}}',
+      {{/if}}
+      {{/each}}
+    ] as const;
```
vmarta, line 22: Example result of rendering this template:

```ts
      static readonly getCategoriesParamKeys = [
        'organizationId',
        'ids',
        'levels',
        'locale',
        'siteId',
      ] as const;
  
      static readonly getCategoriesParamKeysRequired = [
        'organizationId',
        'ids',
        'siteId',
      ] as const;
```

package.json
```diff
@@ -177,11 +177,11 @@
   "bundlesize": [
     {
       "path": "lib/**/*.js",
-      "maxSize": "46 kB"
+      "maxSize": "48 kB"
```
joeluong-sfcc, line 180: I think there's an opportunity to refactor so that we don't have to increase the max bundle size. If you take a look at the rendered templates and look at the generated typescript code for setting query parameters, we have an if/else statement for each query parameter. For example, this is what it looks like for `getProduct`:

```javascript
...
if (optionParams["select"] !== undefined) {
          queryParams["select"] = optionParams["select"];
        } else if (configParams["select"] !== undefined) {
          queryParams["select"] = configParams["select"];
        }
        if (optionParams["inventoryIds"] !== undefined) {
          queryParams["inventoryIds"] = optionParams["inventoryIds"];
        } else if (configParams["inventoryIds"] !== undefined) {
          queryParams["inventoryIds"] = configParams["inventoryIds"];
        }
        if (optionParams["currency"] !== undefined) {
          queryParams["currency"] = optionParams["currency"];
        } else if (configParams["currency"] !== undefined) {
          queryParams["currency"] = configParams["currency"];
        }
        if (optionParams["expand"] !== undefined) {
          queryParams["expand"] = optionParams["expand"];
        } else if (configParams["expand"] !== undefined) {
          queryParams["expand"] = configParams["expand"];
        }
        if (optionParams["locale"] !== undefined) {
          queryParams["locale"] = optionParams["locale"];
        } else if (configParams["locale"] !== undefined) {
          queryParams["locale"] = configParams["locale"];
        }
        if (optionParams["allImages"] !== undefined) {
          queryParams["allImages"] = optionParams["allImages"];
        } else if (configParams["allImages"] !== undefined) {
          queryParams["allImages"] = configParams["allImages"];
        }
...
```

But now that we expose the `ParamKeys`, we could theoretically do something like:

```javascript
getProductParamKeys.forEach(key => {
        if (optionParams[key] !== undefined) {
          queryParams[key] = optionParams[key];
        } else if (configParams[key] !== undefined) {
          queryParams[key] = configParams[key];
        }
})
```
There's probably going to be additional refactoring to ensure the proper separation of query and path parameters, but this approach should help reduce the current bundle size

https://github.com/SalesforceCommerceCloud/commerce-sdk-isomorphic/blob/778305483021d5daeaaa5af71f332b8c5e3ec6b2/templates/operations.ts.hbs#L146

vmarta: @joeluong-sfcc good catch! Let me give that a try 👀

alexvuong (approved):

package.json
vmarta: I've tried and gone down the rabbit hole. But it turns out that the resulting _gzip_ size is very similar to before. It looks like gzip compression does a very good job already.

vmarta: So I'm thinking to revert back to 467bf49ff756efde2274d067222370fd3f123a83. It has much smaller change, compared to what I have now.

templates/client.ts.hbs
```diff
@@ -102,5 +102,13 @@ export class {{name.upperCamelCase}}<ConfigParameters extends {{{name.upperCamel
   }
 
   {{> operationsPartial model}}
+}
 
+const throwMissingRequiredParamError = ({name, paramType}: {name: string, paramType: 'query' | 'path'}) => {
+  throw new Error(`Missing required ${paramType} parameter: ${name}`);
```
joeluong-sfcc, line 108: Can we also include the method name with the error for better context?
```suggestion
  throw new Error(`Missing required ${paramType} parameter for ${methodName}: ${name}`);
```

Additionally, we should consider pulling these out into a helper file and importing them to use instead of defining them in the template

templates/operations.ts.hbs
```diff
@@ -129,38 +130,40 @@
...
-        throw new Error('Missing required query parameter: {{{name}}}');
-      }
-      {{/if}}
-      {{/each}}
+      {{{name}}}QueryParamKeys.forEach((paramKey) => {
+        const value = optionParams[paramKey] ?? configParams[paramKey];
+
+        if (value !== undefined) {
+          // @ts-ignore
+          queryParams[paramKey] = value;
```
joeluong-sfcc, line 152: More verbose, but if you want to get rid of the `// @ts-ignore` comment
```suggestion
          queryParams[paramKey as keyof typeof queryParams] = value;
```

joeluong-sfcc (approved):

kevinxh:
In the `commerce-sdk-react` PR, we have 

```js
import {ShopperProductsParamKeys as paramKeys} from 'commerce-sdk-isomorphic'
export default {
    getProducts: paramKeys.getProducts,
    getProductsRequired: paramKeys.getProductsRequired,

    getProduct: paramKeys.getProduct,
    getProductRequired: paramKeys.getProductRequired,

    getCategories: paramKeys.getCategories,
    getCategoriesRequired: paramKeys.getCategoriesRequired,

    getCategory: paramKeys.getCategory,
    getCategoryRequired: paramKeys.getCategoryRequired
}
```

as we can see, we no longer need to hard code the parameters, yay! 

But, we still need to hard code the endpoints like `getProduct`...

I wonder if we can export the list of available endpoint from the isomorphic library? That way we move one step closer to automatically generate the react lib.

templates/index.ts.hbs
```diff
@@ -1,5 +1,5 @@
 {{#each children}}
-export { {{name.upperCamelCase}} } from "./{{name.lowerCamelCase}}";
+export { {{name.upperCamelCase}}, paramKeys as {{name.upperCamelCase}}ParamKeys } from "./{{name.lowerCamelCase}}";
```
adamraya, line 2: I wonder if a better interface would be to provide `ShopperProductsParamKeys` and `ShopperProductsParamKeysRequired` so it's easier to consume the types and avoid using template strings in the commerce-sdk-react like:

```
paramKeysMap[`${methodName}Required`]
```

vmarta:
> I wonder if we can export the list of available endpoint from the isomorphic library? That way we move one step closer to automatically generate the react lib.

@kevinxh yes that's possible. We can use the `operations` property in the template to loop over the endpoints.

templates/index.ts.hbs
vmarta: The template string above is actually not necessary. I was following the pattern that already exists in the file. But we can access the parameter keys directly if we want to.

Issue #159, opened 2024-06-06
[Feature] Allow providing your own implementation of `fetch`


johnboxall:
Application Performance Monitoring frameworks often want to spy on HTTP requests being made.

To to this, they typically need to monkey patch / override `fetch`.

Currently its tricky to do this:

https://sfcc-unofficial.slack.com/archives/C01AWMT7GSZ/p1717714138887519

Ideally, we'd provide some way of providing `fetch` implementation, potentially as an additional param when creating an API client.

clavery:
This would also be useful for general request/response interception use cases.

opepin:
followed @clavery  suggestion 

`var myFetch: typeof globalObject.fetch;
  return function fetch() {
    if (!myFetch) {
      myFetch = globalObject.fetch;
    }
    // @ts-ignore
    return myFetch.apply(globalObject, arguments)
  }`
  
  Fetched can is now instrumented as expected and addressees missing requests  https://docs.logrocket.com/docs/troubleshooting-sessions#network-requests-are-missing-or-not-being-recorded

PR #160, opened 2024-06-08
Issue 159

opepin:
Allow providing your own implementation of fetch by delaying caching from instanciation to first call. 

Fetch can now be instrumented for performance monitoring/error or data capture. https://docs.logrocket.com/docs/troubleshooting-sessions#network-requests-are-missing-or-not-being-recorded


joeluong-sfcc:
Hi @opepin, thanks for creating this PR. Would you be willing to provide some documentation/example on how you would pass in a custom implementation of a fetch call and adding a unit test?

opepin:
I don't have a unit test, but the idea would be to override the window.fetch method : 

const { fetch: originalFetch } = window;

window.fetch = async (...args) => {
    let [resource, config ] = args;
    console.log('request interceptor before')
    const response = await originalFetch(resource, config);
    console.log('request interceptor after')
    return response;
};

before calling the api

opepin:
fixes #159

PR #162, opened 2024-06-18
Release v2.1.0 (@W-15945047@)

vmarta:
This PR plans to release the following items:
- Exported the parameter keys for each API endpoint
- Added a new endpoint `updateItemsInBasket`


.github/workflows/test.yml
```diff
@@ -9,14 +9,14 @@ jobs:
   linux-tests:
     strategy:
       matrix:
-        node: [10, 12, 14, 16, 18, 20]
+        node: [12, 14, 16, 18]
```
vmarta, line 12: Removed Node 20 for now because it was never working (i.e. you can't build/develop it successfully, but developers should still be able to use it in their project). 

CI did not catch this error because of incorrect setup. See below.

.github/workflows/test.yml
```diff
@@ -9,14 +9,14 @@ jobs:
...
       fail-fast: false
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
       - uses: actions/setup-node@v3
         with:
-          node-version: ${{ matrix.node_version }}
+          node-version: ${{ matrix.node }}
```
vmarta, line 19: Fixed the CI tests to make sure that each environment uses the correct Node version. They were all using the same Node 18 because of the typo.

package.json
```diff
@@ -181,7 +181,7 @@
     },
     {
       "path": "commerce-sdk-isomorphic-with-deps.tgz",
-      "maxSize": "410 kB"
+      "maxSize": "425 kB"
```
vmarta, line 184: Had to increase this bundle size limit because different Node versions actually result in different bundle size.

.nvmrc
```diff
@@ -1 +1 @@
-10
\ No newline at end of file
+14
```
joeluong-sfcc, line 1: What does this change do?

joeluong-sfcc (approved):

.github/workflows/test.yml
joeluong-sfcc: I think we should create a ticket/github issue to track the work to eventually add this back in

.nvmrc
vmarta: Oh I wanted to change the default Node version (for `nvm`) to be Node 14, matching what the `publish` workflow uses.

.github/workflows/test.yml
vmarta: 👍 ok I'll create the ticket.

joeluong-sfcc (approved):

unandyala (approved):

alexvuong (approved):

PR #165, opened 2024-07-15
Update SLAS helper to require channel_id

joeluong-sfcc:
Starting 7/31/2024, SLAS will start requiring the `channel_id` as a required argument when retrieving an access token with `grant_type: 'client_credentials` for security purposes. This PR updates the `loginGuestUserPrivate` SLAS helper function to require `channel_id`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # CHANGELOG
 
+## v3.0.0
+
+### Enchancements
+
+- Update SLAS helper function `loginGuestUserPrivate` to require `channel_id` [#165](https://github.com/SalesforceCommerceCloud/commerce-sdk-isomorphic/pull/165)
+
 ## v2.1.0
 
 ### Enhancements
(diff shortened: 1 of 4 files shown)


unandyala (approved):

unandyala (approved):

adamraya (commented):
Not related to this change but I see the `loginGuestUserPrivate` is async, but I don't see any `await` statement inside the function.


Are we missing an `await` statement here:
https://github.com/SalesforceCommerceCloud/commerce-sdk-isomorphic/blob/ba0e1a9765879326190e5d9904e5f74e0c4982ce/src/static/helpers/slasHelper.ts#L206

```
return await slasClient.getAccessToken(options);
```

adamraya (approved):

adamraya:
> Someone can correct my javascript knowledge if its incorrect, but I believe that marking a function with async means that return value is automatically a promise, so although we don't use `await` statements, we return a `Promise<TokenResponse>`

Yes you're right, the function marked with async automatically returns a promise. It will not wait for getAccessToken to resolve before returning, but I guess subsequent operations are handled in getAccessToken.

PR #167, opened 2024-08-08
Add support for SLAS DNT (do not track) parameter

kevinxh:
This PR adds a new optional parameter `dnt` to SLAS helpers: `loginGuestUserPrivate`, `loginGuestUser` and `loginRegisteredUserB2C`

### DIFF
--- a/src/static/helpers/slasHelper.test.ts
+++ b/src/static/helpers/slasHelper.test.ts
@@ -1,7 +1,7 @@
 /**
  * @jest-environment node
  */
-/* eslint header/header: "off" */
+/* eslint header/header: "off", max-lines:"off" */
 /*
  * Copyright (c) 2022, salesforce.com, inc.
  * All rights reserved.
@@ -47,6 +47,7 @@ const parameters = {
   refreshToken: 'refresh_token',
   usid: 'usid',
   hint: 'hint',
+  dnt: 'true',
 };
 
 const url =
@@ -55,9 +56,7 @@ const url =
 const authenticateCustomerMock = jest.fn(() => ({url}));
 
 const getAccessTokenMock = jest.fn(() => expectedTokenResponse);
-
 const logoutCustomerMock = jest.fn(() => expectedTokenResponse);
-
 const generateCodeChallengeMock = jest.fn(() => 'code_challenge');
 
 const createMockSlasClient = () =>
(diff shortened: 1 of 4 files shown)


src/static/helpers/slasHelper.test.ts
```diff
@@ -47,6 +47,7 @@ const parameters = {
   refreshToken: 'refresh_token',
   usid: 'usid',
   hint: 'hint',
+  dnt: 'true',
```
kevinxh, line 50: well, the API spec says this parameter is `string`.. which is kinda weird to me, shouldn't it be a `boolean`? I'm gonna ask

joeluong-sfcc: Agreed that it is a bit weird, but if it changes in the RAML spec from type `string` to type `boolean` it'll be a breaking change 😭

joeluong-sfcc (commented):
Can you update the CHANGELOG? otherwise LGTM

src/static/helpers/slasHelper.test.ts
joeluong-sfcc: Agreed that it is a bit weird, but if it changes in the RAML spec from type string to type boolean it'll be a breaking change :(

kevinxh: Good call. I'm going to make the SLAS helpers to accepts booleans and do a `.toString()` to convert it before calling the API client.

src/static/helpers/slasHelper.ts
```diff
@@ -166,6 +166,7 @@ export async function authorize(
  * @param credentials.clientSecret - secret associated with client ID
  * @param parameters - parameters to pass in the API calls.
  * @param parameters.usid? - Unique Shopper Identifier to enable personalization.
+ * @param parameters.dnt? - Optional parameter to enable Do Not Track (DNT) for the user.
```
unandyala, line 169: Since it is boolean, does it make sense to set default value instead of making the parameter optional?`dnt=true`

joeluong-sfcc (commented):
LGTM, but one thing we should be weary of is if this `dnt` field will be subject to change in the future. If it gets updated to accept other string values besides `'true'` and `'false'` this implementation will break

If we expect this to be updated in the future we should probably revert to the previous implementation. Given the nature of the field I can't imagine any other options, but better to be safe than sorry, let's check with the SLAS team first if we want to use this approach

src/static/helpers/slasHelper.ts
kevinxh: the default is controlled at the API level, if this parameter is not present, SLAS default to `dnt=false`.

joeluong-sfcc (approved):

unandyala (approved):

PR #173, opened 2024-09-18
@W-16553557 Add passwordless login helpers

jeremy-jung1:
Adding login helpers in isomorphic for passwordless login

E2E Test:
1. Go to https://runtime-admin-n18.mobify-storefront.com/
2. Open Network tab, and put "passwordless" in the search bar
3. Refresh. See a 200 response from `passwordless/login`.
4. Go to https://webhook.site/#!/view/4cb9f313-a8e0-4066-bf2e-5087edf1bf66/5abe16c4-b112-4e12-90cd-8e91aea88598/1
5. Click on the latest POST request, and copy the "token" field in the shown content of the request.
6. Go back to the tab with the storefront. Scroll all the way down until you see a small submission form at the bottom left of the bottom of the window (should say "Enter text")
7. Have the network tab open, and paste the "token" in the input field and press submit. 
8. See 200 response from 'passwordless/token`


src/static/helpers/passwordlessHelper.ts
```diff
@@ -0,0 +1,116 @@
...
+  }
+): Promise<TokenResponse> {
+  const codeVerifier = createCodeVerifier();
+  const authHeaderIdSecret = `Basic ${stringToBase64(
+    `${slasClient.clientConfig.parameters.clientId}:${credentials.clientSecret}`
+  )}`;
+
+  const tokenBody = {
+    grant_type: 'client_credentials',
+    hint: 'pwdless_login',
```
jeremy-jung1, line 102: These parameters come from the API docs: https://developer.salesforce.com/docs/commerce/commerce-api/references/shopper-login?meta=getPasswordLessAccessToken

src/static/helpers/index.ts
```diff
@@ -8,6 +8,7 @@
 // Doing so may lead to circular dependencies or duplicate exports (due to rollup mangling the types)
 export * from './environment';
 export * from './slasHelper';
+export * from './passwordlessHelper';
```
bendvc, line 11: NIT: The passwordless login API's are part of SLAS right? And from the looks of it, the content of this file is somewhat small, is there a reason you didn't slap it in the slasHelpers, sounds like that is the right context for it.

src/static/helpers/passwordlessHelper.ts
```diff
@@ -0,0 +1,116 @@
...
+    locale?: string;
+    mode: string;
+  }
+): Promise<Response> {
+  const authHeaderIdSecret = `Basic ${stringToBase64(
+    `${slasClient.clientConfig.parameters.clientId}:${credentials.clientSecret}`
+  )}`;
+  const tokenBody = {
+    user_id: parameters.userid,
+    mode: parameters.mode,
```
bendvc, line 47: NIT: It doesn't look like we have have a pattern for throwing when required params are not included, so I don't expect you to add it now because I like consistency more. But I'll take note on that.

src/static/helpers/passwordlessHelper.ts
```diff
@@ -0,0 +1,116 @@
...
+ */
+export async function authorizePasswordless(
+  slasClient: ShopperLogin<{
+    shortCode: string;
+    organizationId: string;
+    clientId: string;
+    siteId: string;
+  }>,
+  credentials: {
+    clientSecret: string;
```
bendvc, line 32: You might want to throw if this information is not provided as we are doing [here](https://github.com/SalesforceCommerceCloud/commerce-sdk-isomorphic/blob/10a7405a6594faed7d63a97bc64acc3769003375/src/static/helpers/slasHelper.ts#L187C3-L191C4). This comment also applies to the `getPasswordLessAccessToken` function implementation below.

src/static/helpers/passwordlessHelper.ts
jeremy-jung1: It also seems like typescript checks for the existence of these params that are directly required. But when it's parameters within an object like ShopperLogin, typescript doesn't automatically give an error that expected arguments don't exist

bendvc: We shouldn't reply on TypeScript types for things like that because this lib is compiled into JS and it can be used by a project that isn't TS. Typescript won't bake in any kind of parameter checking so there is the possibility that you this code would make a server request that would for sure fail if that param is required. 

But again, no need to change that now, it doesn't seem like it's a pattern that we have, and we should probably fix that in another PR in one big swoop.

src/static/helpers/index.ts
bendvc: Thats super lame, seems a little arbitrary. 

It definitely seems like the SLAS passwordless helper should be located along with all the other SLAS helpers. Can you propose an alternative (non-breaking) solution here?

joeluong-sfcc: Agreed that it's pretty arbitrary, when the line limit was implemented (3 years ago) I don't think the team originally envisioned a helper file growing to this size. I think it'd be ok to update the line limit or ignore it for this file:

Ignore: `/* eslint-disable max-lines */`
Update max lines: https://github.com/SalesforceCommerceCloud/commerce-sdk-isomorphic/blob/2af2671a69b6e7b656d91cbc5b319cd1066eeb33/.eslintrc.json#L57

jeremy-jung1: I'll remove the maxlines to make it as flexible as possible for new helpers as part of this epic

bendvc (approved):

yunakim714 (approved):
E2e tests working as expected and the helpers look good to me!

PR #178, opened 2025-01-13
Update CI to use node 22 @W-16170810@

joeluong-sfcc:
This PR updates the publish and test workflows to use node 22

This PR also switches the `bundlesize` package for the `bundlesize2` package as `yarn install` was failing due to a transitive dependency `iltorb` in `bundlesize` that is not supported in newer node versions.

There will be a follow up PR in the future to remove old node versions from the CI when we drop official support for them.

NOTE: we're holding off on merging this until bundlesize2 is approved as a 3PP

UPDATE: bundlesize2 has been approved


.github/workflows/test.yml
```diff
@@ -9,7 +9,8 @@ jobs:
   linux-tests:
     strategy:
       matrix:
-        node: [12, 14, 16, 18]
+        # TODO: remove older versions of node
+        node: [12, 14, 16, 18, 20, 22]
```
unandyala, line 13: Just curious - Is there a specific reason why we want to remove older node versions ins a separate PR?

unandyala (approved):
LGTM

.github/workflows/test.yml
joeluong-sfcc: Yes, the plan is to remove the older node versions when we drop official support for them, which is a breaking change, so we plan to remove them in the next major version most likely when we release the OAS SDK

alexvuong (approved):

PR #186, opened 2025-03-21
W-18089100 Allow custom params in helpers

alexvuong:
This PR allows custom parameters to be passed into `/authorize` endpoint via the `loginGuestUser` and `authorizeIDP` helpers  and `/login` via `loginRegisteredUsers` helper. 

- Customer headers are not allowed and will be ignored via SLAS
Some Note on the authorize endpoint
- The /authorize endpoint does not call SCAPI / OCAPI customers/auth where the hooks take place.
- The /authorize endpoint is used for "Social" or 3rd party logins (okta, Auth0, etc.). SLAS will call out to the given Identity Provider (IDP) to have the user log in.
- For /authorize endpoint the "c_" parameters are passed to the 3rd party IDP and no ECOM hooks are involved.
- /login will have authentication hooks. See [here](https://developer.salesforce.com/docs/commerce/commerce-api/guide/hook-method-details.html)

How to test drive PR
- Check out code
- yarn install
- yarn test and confirm all the tests are all passed

### DIFF
--- a/src/static/helpers/slasHelper.ts
+++ b/src/static/helpers/slasHelper.ts
@@ -109,7 +109,7 @@ export const generateCodeChallenge = async (
  * @param parameters.redirectURI - the location the client will be returned to after successful login with 3rd party IDP. Must be registered in SLAS.
  * @param parameters.hint? - optional string to hint at a particular IDP. Guest sessions are created by setting this to 'guest'
  * @param parameters.usid? - optional saved SLAS user id to link the new session to a previous session
- * @param privateClient - flag to indicate if the client is private or not. Defaults to false.
+ * @param privateClient? - flag to indicate if the client is private or not. Defaults to false.
  * @param options - an object containing the options for this method.
  * @param options.headers? - optional header to pass in the 'authorizeCustomer` endpoint.
  * @returns login url, user id and authorization code if available
--- a/src/static/helpers/slasHelper.ts
+++ b/src/static/helpers/slasHelper.ts
@@ -19,7 +19,11 @@ import {
 } from '../../lib/shopperLogin';
 import ResponseError from '../responseError';
 import TemplateURL from '../templateUrl';
-import {BaseUriParameters} from './types';
+import {
+  BaseUriParameters,
+  CustomQueryParameters,
+  CustomRequestBody,
+} from './types';
 
 export const stringToBase64 = isBrowser
   ? btoa
@@ -106,7 +110,8 @@ export const generateCodeChallenge = async (
  * @param parameters.hint? - optional string to hint at a particular IDP. Guest sessions are created by setting this to 'guest'
  * @param parameters.usid? - optional saved SLAS user id to link the new session to a previous session
  * @param privateClient - flag to indicate if the client is private or not. Defaults to false.
- * @param headers - optional headers to pass in the 'authorizeCustomer` endpoint.
+ * @param options - an object containing the options for this method.
+ * @param options.headers? - optional header to pass in the 'authorizeCustomer` endpoint.
  * @returns login url, user id and authorization code if available
  */
 export async function authorize(
@@ -121,9 +126,11 @@ export async function authorize(
     redirectURI: string;
     hint?: string;
     usid?: string;
-  } & {[key in `c_${string}`]: any},
+  } & CustomQueryParameters,
   privateClient = false,
-  headers?: {[key: string]: string}
+  options?: {
+    headers?: {[key: string]: string};
+  }
 ): Promise<{code: string; url: string; usid: string}> {
   interface ClientOptions {
     codeChallenge?: string;
@@ -145,25 +152,25 @@ export async function authorize(
     redirect: isBrowser ? 'follow' : 'manual',
   };
 
-  const options = {
+  const {hint, redirectURI, usid, ...restOfParams} = parameters;
+  const opts = {
     parameters: {
       client_id: slasClient.clientConfig.parameters.clientId,
       channel_id: slasClient.clientConfig.parameters.siteId,
       ...(clientOptions.codeChallenge && {
         code_challenge: clientOptions.codeChallenge,
       }),
-      ...(parameters.hint && {hint: parameters.hint}),
+      ...(hint && {hint}),
       organizationId: slasClient.clientConfig.parameters.organizationId,
-      redirect_uri: parameters.redirectURI,
+      redirect_uri: redirectURI,
       response_type: 'code',
-      ...(parameters.usid && {usid: parameters.usid}),
-      // we don't need to validate c_params because shopperLogin func will do that
-      ...parameters,
+      ...(usid && {usid}),
+      ...restOfParams,
     },
-    headers,
+    ...(options?.headers || {}),
   };
 
-  const response = await slasClientCopy.authorizeCustomer(options, true);
+  const response = await slasClientCopy.authorizeCustomer(opts, true);
   const redirectUrlString = response.headers?.get('location') || response.url;
   const redirectUrl = new URL(redirectUrlString);
   const searchParams = Object.fromEntries(redirectUrl.searchParams.entries());
@@ -254,7 +261,8 @@ export async function authorizeIDP(
  * @param parameters.redirectURI - Per OAuth standard, a valid app route. Must be listed in your SLAS configuration. On server, this will not be actually called. On browser, this will be called, but ignored.
  * @param parameters.usid? - Unique Shopper Identifier to enable personalization.
  * @param parameters.dnt? - Optional parameter to enable Do Not Track (DNT) for the user.
- * @param headers - optional headers to pass in the 'getAccessToken'
+ * @param options - an object containing the options for this method
+ * @param options?.headers - optional headers to pass in the 'getAccessToken'
  * @returns TokenResponse
  */
 export async function loginIDPUser(
@@ -273,8 +281,10 @@ export async function loginIDPUser(
     code: string;
     usid?: string;
     dnt?: boolean;
-  } & {[key in `c_${string}`]: any},
-  headers?: {[key: string]: string}
+  } & CustomQueryParameters,
+  options?: {
+    headers?: {[key: string]: string};
+  }
 ): Promise<TokenResponse> {
   const privateClient = !!credentials.clientSecret;
   const {code, dnt, usid, redirectURI, ...restOfParams} = parameters;
@@ -304,14 +314,17 @@ export async function loginIDPUser(
     const optionsToken = {
       headers: {
         Authorization: authHeaderIdSecret,
-        ...headers,
+        ...options?.headers,
       },
       body: tokenBody,
     };
     return slasClient.getAccessToken(optionsToken);
   }
   // default is to use slas public client
-  return slasClient.getAccessToken({body: tokenBody, headers});
+  return slasClient.getAccessToken({
+    body: tokenBody,
+    headers: options?.headers,
+  });
 }
 
 /**
@@ -323,7 +336,8 @@ export async function loginIDPUser(
  * @param parameters - parameters to pass in the API calls.
  * @param parameters.usid? - Unique Shopper Identifier to enable personalization.
  * @param parameters.dnt? - Optional parameter to enable Do Not Track (DNT) for the user.
- * @param headers - optional headers to pass in the 'getAccessToken` endpoint.
+ * @param options - an object containing the options for this method.
+ * @param option.headers? - optional headers to pass in the 'getAccessToken` endpoint.
  * @returns TokenResponse
  */
 export async function loginGuestUserPrivate(
@@ -336,11 +350,13 @@ export async function loginGuestUserPrivate(
   parameters: {
     usid?: string;
     dnt?: boolean;
-  } & {[key in `c_${string}`]: any},
+  } & CustomQueryParameters,
   credentials: {
     clientSecret: string;
   },
-  headers?: {[key: string]: string}
+  options?: {
+    headers?: {[key: string]: string};
+  }
 ): Promise<TokenResponse> {
   if (!slasClient.clientConfig.parameters.siteId) {
     throw new Error(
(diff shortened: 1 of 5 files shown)


src/static/helpers/slasHelper.ts
```diff
@@ -426,6 +454,10 @@ export async function loginRegisteredUserB2C(
     redirectURI: string;
     usid?: string;
     dnt?: boolean;
+  } & {[key in `c_${string}`]: any},
+  opts?: {
+    headers?: {[key: string]: string};
+    body?: {[key in `c_${string}`]: any} & LoginRequest;
```
alexvuong, line 460: Reason we allow body here is the `authenticateCustomer` [API](https://developer.salesforce.com/docs/commerce/commerce-api/references/shopper-login?meta=authenticateCustomer) nows allows a custom body that has pattern ^c_.+$. We should allow that to go through slas helper here.

alexvuong, line 460: Reason we allow body here is the `authenticateCustomer` [API](https://developer.salesforce.com/docs/commerce/commerce-api/references/shopper-login?meta=authenticateCustomer) nows allows a custom body that has pattern ^c_.+$. We should allow that to go through slas helper here.

alexvuong:
The login* functions usually will go through two funcs to get the access. One is to call `authenticate*`  and `getAccessToken`. Would it make sense to pass the arguments `headers` passed into the `login*` to both functions? 🤔

src/static/helpers/slasHelper.ts
```diff
@@ -118,8 +121,9 @@ export async function authorize(
     redirectURI: string;
     hint?: string;
     usid?: string;
-  },
-  privateClient = false
+  } & {[key in `c_${string}`]: any},
```
joeluong-sfcc, line 124: I think we should create a type for custom query parameters and reuse it instead of using `any`:

```typescript
/**
 * Custom query parameter type with any string prefixed with `c_` as the key and the allowed 
 * types for query parameters for the value.
 */
export type CustomQueryParameters = {
  [key in `c_${string}`]: string | number | boolean | string[] | number[];
}


// in function
  parameters: {
    redirectURI: string;
    hint?: string;
    usid?: string;
  } & CustomQueryParameters,
```

It'll be a breaking change if we update it in our API files, but in the next major release we can add a note to update it to use that `CustomQueryParameters` type instead of `any` as well

src/static/helpers/slasHelper.ts
```diff
@@ -104,6 +105,8 @@ export const generateCodeChallenge = async (
  * @param parameters.redirectURI - the location the client will be returned to after successful login with 3rd party IDP. Must be registered in SLAS.
  * @param parameters.hint? - optional string to hint at a particular IDP. Guest sessions are created by setting this to 'guest'
  * @param parameters.usid? - optional saved SLAS user id to link the new session to a previous session
+ * @param privateClient - flag to indicate if the client is private or not. Defaults to false.
```
joeluong-sfcc, line 108:
```suggestion
 * @param privateClient? - flag to indicate if the client is private or not. Defaults to false.
```

For this one and the other ones please

src/static/helpers/slasHelper.ts
```diff
@@ -104,6 +105,8 @@ export const generateCodeChallenge = async (
...
+ * @param headers - optional headers to pass in the 'authorizeCustomer` endpoint.
```
joeluong-sfcc, line 109:
```suggestion
 * @param headers? - optional headers to pass in the 'authorizeCustomer` endpoint.
```

For this one and the other ones please

src/static/helpers/slasHelper.ts
```diff
@@ -153,7 +157,10 @@ export async function authorize(
       redirect_uri: parameters.redirectURI,
       response_type: 'code',
       ...(parameters.usid && {usid: parameters.usid}),
+      // we don't need to validate c_params because shopperLogin func will do that
+      ...parameters,
```
joeluong-sfcc, line 161: I think this spread should go at the top, before we start defining properties in `parameters` so that we don't accidentally override the properties we previously defined:

```typescript
    parameters: {
      // we don't need to validate c_params because shopperLogin func will do that
      ...parameters,
      client_id: slasClient.clientConfig.parameters.clientId,
      channel_id: slasClient.clientConfig.parameters.siteId,
      ...(clientOptions.codeChallenge && {
        code_challenge: clientOptions.codeChallenge,
      }),
      ...(parameters.hint && {hint: parameters.hint}),
      organizationId: slasClient.clientConfig.parameters.organizationId,
      redirect_uri: parameters.redirectURI,
      response_type: 'code',
      ...(parameters.usid && {usid: parameters.usid}),
   },
```

Let's update the other functions as well

src/static/helpers/slasHelper.ts
joeluong-sfcc: Now that the SLAS helpers have gotten kind of complicated with numerous arguments, ideally we have all of the arguments under the a single argument called `options` to keep consistent with how the API classes work:

```typescript
loginRegisteredUserB2C(options: {
   slasClient: { ... },
   credentials: { ... },
   parameters: { ... },
   headers?: { ... },
   body?: { ... },
}
```

Unfortunately, this is a breaking change and we'll have to wait till the next major version release to implement something like this.

For now, I do like the idea of the `opts` object, but its inconsistent with the rest of the SLAS helpers. We should either update all the helpers to have an `opts` object where we pass headers through, or we just have `headers` and `body` as separate arguments. If we go with the `opts` object I'd vote to rename it as `options` just for consistency with the API classes.

joeluong-sfcc (changes requested):

alexvuong:
> > The login* functions usually will go through two funcs to get the access. One is to call `authenticate*` and `getAccessToken`. Would it make sense to pass the arguments `headers` passed into the `login*` to both functions? 🤔
> 
> Ah that's kind of annoying, I feel like there are a couple ways to get around this but I can't think of clean way. I think we should provide the ability to pass separate headers to each function if required. One way to do it is like so:
> 
> ```ts
>           headers?: {
>             authenticateHeaders?: { [key: string]: string },
>             getAccessTokenHeaders?: { [key: string]: string }
>           } & { [key: string]: string }
> ```
> 
> We can check if `authenticateHeaders` or `getAccessTokenHeaders` exist and pass them to their respective calls, otherwise we can just pass the whole headers object to each one, it feels a little hacky though
> 
> Wonder if @kevinxh has any opinions about this

This may be fine for just headers, I think we have to think about if that also makes sense for parameters and body too. Each of these calls are accepting its own customer params. `authenticateCustomer` accept custom bodies (e.g c_nickname). 
```
authenticateCustomer<T extends boolean>(
        options: RequireParametersUnlessAllAreOptional<{
          parameters?: CompositeParameters<{
            organizationId: string
          } & { [key in `c_${string}`]: any }, ConfigParameters>,
          headers?: { [key: string]: string },
          body: LoginRequest
        }>,
        rawResponse?: T
      ): Promise<T extends true ? Response : void>;

getAccessToken(
        options: RequireParametersUnlessAllAreOptional<{
          parameters?: CompositeParameters<{
            organizationId: string
          } & { [key in `c_${string}`]: any }, ConfigParameters>,
          headers?: { [key: string]: string },
          body: TokenRequest
        }>
      ): Promise<TokenResponse>;

```

src/static/helpers/slasHelper.ts
```diff
@@ -104,6 +109,9 @@ export const generateCodeChallenge = async (
  * @param parameters.redirectURI - the location the client will be returned to after successful login with 3rd party IDP. Must be registered in SLAS.
  * @param parameters.hint? - optional string to hint at a particular IDP. Guest sessions are created by setting this to 'guest'
  * @param parameters.usid? - optional saved SLAS user id to link the new session to a previous session
+ * @param privateClient? - flag to indicate if the client is private or not. Defaults to false.
+ * @param options - an object containing the options for this function.
```
joeluong-sfcc, line 113:
```suggestion
 * @param options? - an object containing the options for this function.
```

`options` is optional

src/static/helpers/slasHelper.ts
```diff
@@ -141,22 +152,25 @@ export async function authorize(
...
       }),
-      ...(parameters.hint && {hint: parameters.hint}),
+      ...(hint && {hint}),
       organizationId: slasClient.clientConfig.parameters.organizationId,
-      redirect_uri: parameters.redirectURI,
+      redirect_uri: redirectURI,
       response_type: 'code',
-      ...(parameters.usid && {usid: parameters.usid}),
+      ...(usid && {usid}),
+      ...restOfParams,
```
joeluong-sfcc, line 168: Is there a use case where `...restOfParams` should be last instead of first? In my opinion, `..restOfParams` should go on top before we start defining the other parameter properties. Imagine a developer that uses javascript instead of typescript and so they're able to define `codeChallenge: '<insert-custom-string>` in the parameters object they pass in. That would break implementation and so we probably don't want to allow the ability to override these properties already set by us, which is a breaking change

src/static/helpers/slasHelper.ts
```diff
@@ -141,22 +152,25 @@ export async function authorize(
...
     },
+    ...(options?.headers || {}),
```
joeluong-sfcc, line 170: headers have to be defined under the `headers` property
```suggestion
    ...(options?.headers && { headers: options.headers }),
```

src/static/helpers/slasHelper.ts
```diff
@@ -292,13 +314,17 @@ export async function loginIDPUser(
     const optionsToken = {
       headers: {
         Authorization: authHeaderIdSecret,
+        ...options?.headers,
```
joeluong-sfcc, line 317: I'm ok having this the headers spread here since `options?.headers` is a new argument, but we'll allow the possibility for developers to accidentally/incorrectly override `Authorization`

src/static/helpers/slasHelper.ts
alexvuong: the name options is already an indicator that it is optional. I don't think we need to do that

src/static/helpers/slasHelper.ts
kevinxh: what's the use case of overriding the `Authorization` header?

If you want to use a different clientid/secret, you should update the argument for the api client, not overriding it in slas helpers?

If you want to use a different format other than base64 client:secret encoding, does the API still work? If no, we should not allow it to be overriden.

src/static/helpers/slasHelper.ts
kevinxh: If the tests didn't catch this, then we have a problem with the tests...

kevinxh:
> The login* functions usually will go through two funcs to get the access. One is to call `authenticate*` and `getAccessToken`. Would it make sense to pass the arguments `headers` passed into the `login*` to both functions? 🤔

I think it's fine to pass headers to both calls, if and only if SLAS API doesn't reject unknown headers.

This is a trade-off we need to make, helper ergonomics/usability v.s. customizability. There is always a way out for users that need the absolute fine grain customizability - **don't use the helpers!** 

Helpers are there to make everyone's (99% of people) life easier. It defeats the purpose of having helpers if the function signature is so complicated?

That's just my 2 cents

src/static/helpers/slasHelper.ts
joeluong-sfcc: It's minor, but I'm pushing back that we should mark it as optional with `?` to maintain consistency in this file as all other optional arguments are marked with `?` in the TS/JSDocs

I'm also ok updating the SDK to use the square bracket notation to indicate optionality, but we should be consistent in the convention that we use

src/static/helpers/slasHelper.ts
alexvuong: This is my mistake. They will treat these params as custom params for parameters objects, and it is supposed to throw a warning here if there are an invalid params. I added a test for that. Thanks for pointing this out!

src/static/helpers/slasHelper.ts
alexvuong: afaik, there is no other way then the base64 for authorization header. I've moved it above to avoid overriding for it. Thanks for the feedback

joeluong-sfcc (approved):
LGTM, thanks for adding extensive unit testing, also manually smoke tested that I can pass custom params/headers for SLAS helpers, at least for `loginGuestUser`

src/static/helpers/types.ts
```diff
@@ -59,3 +59,19 @@ export interface QueryParameters {
...
+export type CustomQueryParameters = {
+  [key in `c_${string}`]: string | number | boolean | string[] | number[];
+};
+
+/**
+ * Custom body request type with any string prefixed with `c_` as the key and the allowed
+ * types for the value.
+ */
+export type CustomRequestBody = {
+  [key in `c_${string}`]: string | number | boolean | string[] | number[];
```
joeluong-sfcc, line 76: EDIT: sorry I missed this, we should update this type as we can pass in complex objects into the request body, I believe something like this should fix our use case

```suggestion
  [key in `c_${string}`]: string | number | boolean | string[] | number[] | { [key: string]: any };
```

joeluong-sfcc (changes requested):
EDIT: sorry I missed this one point

src/static/helpers/slasHelper.ts
```diff
@@ -426,6 +450,9 @@ export async function loginRegisteredUserB2C(
     redirectURI: string;
     usid?: string;
     dnt?: boolean;
+  } & CustomQueryParameters,
+  options?: {
+    body?: CustomRequestBody;
```
alexvuong, line 455: We should only expect body to just custom params since the required body is constructed via parameters or clientConfig that does not require users input when calling this function

joeluong-sfcc (commented):
Thanks for reaching out to the SLAS team, just a few minor doc updates

src/static/helpers/types.ts
```diff
@@ -59,3 +59,25 @@ export interface QueryParameters {
...
+ * types for the value.
+ */
+export type CustomRequestBody = {
+  [key in `c_${string}`]:
+    | string
+    | number
+    | boolean
+    | string[]
+    | number[]
+    | {[key: string]: any};
```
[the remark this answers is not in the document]
alexvuong: Done, thanks for the feedback

src/static/helpers/slasHelper.ts
```diff
@@ -118,7 +123,7 @@ export async function authorize(
     redirectURI: string;
     hint?: string;
     usid?: string;
-  },
+  } & CustomQueryParameters,
```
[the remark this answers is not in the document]
alexvuong: Changelog/comment are added

kevinxh (approved):

PR #195, opened 2025-05-27
@W-17701711@ Replace RAML files with OAS

vcua-mobify:
PR for replacing the committed RAML api files inside the `apis` directory with OAS api files


joeluong-sfcc (commented):
I noticed that there's only v2 of shopper baskets, based on the previous spike done for shopper baskets do we want to include v1 as well?

vcua-mobify:
Yes, that's correct. Let's do the baskets v1 addition in a separate PR.

PR #196, opened 2025-05-29
WIP: allow unknown query params

unandyala:


templatesOas/apis.mustache
```diff
@@ -406,11 +406,18 @@ export class {{#lambda.titlecase}}{{#lambda.camelcase}}{{appName}}{{/lambda.came
...
-            if(key.startsWith('c_') && optionParams[key as keyof typeof optionParams] !== undefined) {
-                queryParams[key as keyof typeof queryParams] = optionParams[key as keyof typeof optionParams]
-            } else if(!queryParams.hasOwnProperty(key) && !pathParams.hasOwnProperty(key)) {
-                console.warn(`Invalid Parameter for {{{nickname}}}: ${key}`)
+            const paramValue = optionParams[key as keyof typeof optionParams];
+            if(paramValue === undefined) {
+                return;
             }
+            const isCustomParam = key.startsWith('c_')
+            if(!isCustomParam && ((key in queryParams) || (key in pathParams))) {
```
joeluong-sfcc, line 414: I think we can condense this logic, it's not as readable but since this is duplicated across many endpoints we should aim to keep the bundlesize small

```suggestion
            if(!isCustomParam && !((key in queryParams) || (key in pathParams))) {
                console.warn(`...`)
            }
```

joeluong-sfcc (commented):
Are we planning to create a follow up PR for the node SDK once the OAS work gets merged in?

Other than these points LGTM

templatesOas/apis.mustache
unandyala: good point. But this does not help to reduce the lines of code as we need to add both custom params and the query params that are already not added. May be we should move to a helper

PR #198, opened 2025-06-05
@W-18349405@ Download and generate multiple versions of ShopperBaskets

vcua-mobify:
This PR contains the following changes to support handling both ShopperBaskets v1 and v2:

1) An update to allow the generatre-oas script to look into subdirectories for exchange.json files.

With this change, we can now look for the exchange.json files in the following directory structure:

* shopper-baskets
  * shopper-baskets-v1
    * exchange.json
  * shopper-baskets-v2
    * exchange.json

2) Updates to how information is extracted from exchange.json + handling the API name for both ShopperBasketsV1 and ShopperBasketsV2

3. A fix to handle `-internal` vs `-public` OAS files. The generator will only consider `-internal` files.

4. A template adjustment to handle differences in `ShopperProductQueryParameters` between `getProduct` and `getProducts` (Thanks @joeluong-sfcc for this!)

5. A test update following the change to allow unknown query parameters.

6. Update to api versions + a manual fix to `x-sdk-classname` format.

Testing these changes:
Generate the SDk with `npm run renderTemplates`

Run `npm run test`

Verify that all tests pass and that the generated SDK has no errors (areas to look at would be ShopperBaskets V1 / V2, ShopperProducts, ShopperContexts, and the top level `index.ts`)

### DIFF
--- a/src/static/helpers/slasHelper.test.ts
+++ b/src/static/helpers/slasHelper.test.ts
@@ -171,11 +171,11 @@ describe('Authorize user', () => {
       .query(true)
       .reply(303, {response_body: 'response_body'}, {location: url});
 
-    const authResponse = await slasHelper.authorize(
-      mockSlasClient,
+    const authResponse = await slasHelper.authorize({
+      slasClient: mockSlasClient,
       codeVerifier,
-      parameters
-    );
+      parameters,
+    });
     expect(authResponse).toStrictEqual(expectedAuthResponse);
   });
 
@@ -193,7 +193,11 @@ describe('Authorize user', () => {
       .reply(200);
 
     await expect(
-      slasHelper.authorize(mockSlasClient, codeVerifier, parameters)
+      slasHelper.authorize({
+        slasClient: mockSlasClient,
+        codeVerifier,
+        parameters,
+      })
     ).rejects.toThrow(ResponseError);
   });
 
@@ -229,7 +233,12 @@ describe('Authorize user', () => {
         return [303, {response_body: 'response_body'}, {location: url}];
       });
 
-    await slasHelper.authorize(mockSlasClient, codeVerifier, parameters, true);
+    await slasHelper.authorize({
+      slasClient: mockSlasClient,
+      codeVerifier,
+      parameters,
+      privateClient: true,
+    });
 
     // There should be no code_challenge for private client
     const unexpectedQueryParams = {
@@ -252,9 +261,13 @@ test('throws error on 400 response', async () => {
     .reply(400, {response_body: 'response_body'}, {location: ''});
 
   await expect(
-    slasHelper.authorize(mockSlasClient, codeVerifier, {
-      redirectURI: parameters.redirectURI,
-      usid: parameters.usid,
+    slasHelper.authorize({
+      slasClient: mockSlasClient,
+      codeVerifier,
+      parameters: {
+        redirectURI: parameters.redirectURI,
+        usid: parameters.usid,
+      },
     })
   ).rejects.toThrow(ResponseError);
 });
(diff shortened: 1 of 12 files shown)


scripts/generate-oas.ts
```diff
@@ -57,6 +57,21 @@ export function getAPIDetailsFromExchange(directory: string): ApiSpecDetail {
     const exchangeConfig = fs.readJSONSync(
       exchangePath
     ) as download.ExchangeConfig;
+
+    // Special handling for shopper-baskets-v2
+    if (
+      exchangeConfig.assetId === 'shopper-baskets-oas' &&
```
alexvuong, line 63: Can we be a bit generic for the apis ? If other apis happen to have version 2, we have to come back and manually fix this?

src/static/helpers/slasHelper.test.ts
```diff
@@ -232,18 +232,13 @@ describe('Authorize user', () => {
...
-      accessToken: "access_token",
-      client_id: 'client_id',
-      channel_id: 'site_id',
-      dnt: "false",
-      hint: 'hint',
-      redirect_uri: 'redirect_uri',
-      refreshToken: 'refresh_token',
-      response_type: 'code',
-      usid: 'usid',
+    const unexpectedQueryParams = {
```
alexvuong, line 235: Dumb question, why do we need to change this test?

scripts/utils.ts
```diff
@@ -39,18 +39,14 @@ export async function downloadLatestApis(
...
-  const matchedApis = await download.search(`"${name}"`);
-  if (!(matchedApis?.length > 0)) {
-    throw new Error(`No results in Exchange for '${name}'`);
-  }
-  const api = matchedApis.find(
-    (matchedApi: {assetId: string}) => matchedApi?.assetId === name
+  const matchedApis = await download.search(
+    `"${name}" category:Visibility = "External" category:"SDK Type" = "Commerce"  category:"SDK Type" = "Isomorphic"`,
+    undefined,
+    true
```
alexvuong, line 45: I assumed the search has error handling?

src/static/helpers/slasHelper.test.ts
vcua-mobify: This test fix addresses a change that was made to for allowing any query parameters through: https://github.com/SalesforceCommerceCloud/commerce-sdk-isomorphic/pull/196

Rather than check for expected options, I think it's better we assert that the code challenge wasn't included. That way, if new non-breaking options appear in the future, this test will continue to work.

joeluong-sfcc (approved):
Correct me if I'm wrong, but it seems like the main change is we use a new helper function that recursively goes through the `api` directory and collects all the sub directories with an `exchange.json` file into an array

LGTM

scripts/generate-oas.ts
```diff
@@ -37,10 +37,16 @@ function kebabToCamelCase(str: string): string {
...
-    return 'ShopperOrders';
+function appendVersionIfV2(name: string, version: string): string {
+  if (version === 'V1' || version === 'v1') {
+    return name;
+  }
+  return name + version;
+}
+
+export function resolveApiName(name: string, version: string): string {
+  if (name === 'Shopper Baskets OAS') {
```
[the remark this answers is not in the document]
vcua-mobify: I've modified the function to be more generic when appending V2. But I left the special case for appending V1 to ShopperBaskets specifically for now. We'll need to remove it if ShopperBasketsV1 is reverted to just ShopperBaskets by the IPA team.

scripts/generate-oas.ts
```diff
@@ -37,39 +37,69 @@ function kebabToCamelCase(str: string): string {
...
+  if (name === 'Shopper Seo OAS') {
+    apiName = 'ShopperSEO';
+  } else if (name === 'Shopper Context OAS') {
+    apiName = 'ShopperContexts';
+  } else {
+    apiName = name.replace(/\s+/g, '').replace('OAS', '');
   }
-  return name.replace(/\s+/g, '').replace('OAS', '');
+
+  return version !== 'v1' ? apiName + version.toUpperCase() : apiName;
```
unandyala, line 63: This seem to be opposite of what you are doing for Shopper Baskets? Why do we need this?

vcua-mobify: This was my adjustment to allow for a more generic v2 as per your comment above.

Here, if `apiVersion` in exchange.json is v1, we don't append V1. This results in just the api name being returned. Ie. `ShopperProducts`

If the a`apiVersion` in exchange.json is v2, this returns `ShopperProductsV2`.

ShopperBaskets is a special case where the API team have set `x-sdk-classname` to ShopperBasketsV1 in the `-internal` OAS spec. So the API name returned here has to match that (ie. ShopperBasketsV1).

We don't want to append `V1` to other APIs (ie. we don't want ShopperProductsV1).

unandyala (approved):

PR #201, opened 2025-06-12
Feat: Refactor Isomorphic SDK (@W-18709921@)

joeluong-sfcc:
This PR adds a number of refactors:
- SLAS helpers take in 1 options object instead of multiple positional arguments
- Path parameters are encoded by default
- Form parameters have been refactored into a reusable type
- All endpoints accept `c_` custom properties for request body
- API version as client config property has been removed

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
     "renderTemplates": "PACKAGE_VERSION=$(node -p \"require('./package.json').version\") ts-node --compiler-options '{\"module\": \"commonjs\", \"target\": \"ES6\" }' ./scripts/generate-oas.ts",
     "start": "HTTPS=true react-scripts start",
     "pretest": "yarn run lint && yarn run lint:style && depcheck && yarn run check:size",
-    "test": "yarn run check:types && yarn run test:unit",
+    "test": "yarn run check:types && yarn run test:unit && CI=true yarn run test:react",
     "test:react": "react-scripts test --env=jest-environment-jsdom-sixteen src/environment",
     "test:unit": "jest --coverage --testPathIgnorePatterns node_modules src/environment --silent",
     "updateApis": "ts-node --compiler-options '{\"module\": \"commonjs\", \"target\": \"ES6\" }' ./scripts/updateApis.ts && yarn diffApis"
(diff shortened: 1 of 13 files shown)


package.json
```diff
@@ -39,7 +39,7 @@
     "renderTemplates": "PACKAGE_VERSION=$(node -p \"require('./package.json').version\") ts-node --compiler-options '{\"module\": \"commonjs\", \"target\": \"ES6\" }' ./scripts/generate-oas.ts",
     "start": "HTTPS=true react-scripts start",
     "pretest": "yarn run lint && yarn run lint:style && depcheck && yarn run check:size",
-    "test": "yarn run check:types && yarn run test:unit && CI=true yarn run test:react",
+    "test": "yarn run check:types && yarn run test:unit",
```
vcua-mobify, line 42: Is this an intentional change or a temp change to make running tests faster?
Do we still need to keep a command around where `CI=true` (maybe update test:react)?

joeluong-sfcc: This is an intentional change as the react sample app doesn't work so it doesn't make sense to test it

src/static/templateUrl.test.ts
```diff
@@ -6,6 +6,10 @@
  */
 import TemplateURL from './templateUrl';
 
+// TODO: add unit tests for templateURL
```
unandyala, line 9: May be we can add tests in the PR and rest can be a separate ticket if needed

joeluong-sfcc: Added unit tests!

vcua-mobify (approved):
Thanks for adjusting. LGTM

src/static/helpers/slasHelper.ts
```diff
@@ -110,21 +110,23 @@ export const generateCodeChallenge = async (
...
-export async function authorize(
+export async function authorize(options: {
   slasClient: ShopperLogin<{
     shortCode: string;
     organizationId: string;
     clientId: string;
     siteId: string;
-  }>,
-  codeVerifier: string,
+  }>;
```
alexvuong, line 119: how is this code working, we are changing from colon to semincolon. The colon is to separate the arg in the same function, while the semicoln does not really make sense

alexvuong (approved):

PR #202, opened 2025-06-17
Feat: Support x-scapi-internal (@W-18760264@)

joeluong-sfcc:
This PR adds support for the `x-scapi-internal` custom vendor extension. Whenever an operation is tagged as `x-scapi-internal: true`, we'll hide that operation from being generated in the SDK

Example:
```
    get:
      x-scapi-internal: true
      description: ...
      operationId: getProduct
```

There's unfortunately lots of duplicated code as we have to handle 2 cases

1. The operation has vendor extensions and x-scapi-internal is present
2. The operation has no vendor extensions

```
{{#vendorExtensions}}
{{^x-scapi-internal}}
// Render content for operations that have vendor extensions but NOT x-scapi-internal
{{/x-scapi-internal}}
{{/vendorExtensions}}
{{^vendorExtensions}}
// Render content for operations that have no vendor extensions at all
{{/vendorExtensions}}
```


templatesOas/apis.mustache
```diff
@@ -22,6 +22,8 @@ import type {
 {{#hasEnums}}
 {{#operations}}
 {{#operation}}
+{{#vendorExtensions}}
```
unandyala, line 25: Essentially this PR includes/generates type and operations if
1. there are no vendor extensions
2. there are vendor extensions but no x-scapi-internal extension. Is that correct?

templatesOas/apis.mustache
```diff
@@ -44,12 +46,40 @@ export type {{operationIdCamelCase}}{{enumName}} =
 {{/stringEnums}}
 {{/isEnum}}
 {{/allParams}}
+{{/x-scapi-internal}}
```
unandyala, line 49: Can we include a mustache template in another mustache template. If that is possible may be we can avoid duplicate code

unandyala (approved):

templatesOas/apis.mustache
joeluong-sfcc: Yes, that is correct

vcua-mobify (approved):
I've verified that when `x-scapi-internal` is added to an endpoint, that endpoint does not  get added to the generated SDK.

Endpoints with other custom annotations that don't include `x-scapi-internal` will still appear.

PR #208, opened 2025-08-06
Remove internal OAS file and Fix generation script (@W-19076728@)

joeluong-sfcc:
This PR is waiting for this PR to be merged in and released first: https://github.com/SalesforceCommerceCloud/raml-toolkit/pull/235

This PR fixes the `updateApis` command and the `diffApis` command so that we're able to diff APIs properly. This PR also adds a post processing step where we remove the `internal` OAS file from the `apis/` directory. This is because we'll only be exposing the public OAS file.


scripts/updateApis.ts
```diff
@@ -21,13 +21,44 @@ if (!process.env.ANYPOINT_USERNAME || !process.env.ANYPOINT_PASSWORD) {
 const OLD_APIS_PATH = path.join(__dirname, '../temp/oldApis');
 const PRODUCTION_API_PATH = path.join(__dirname, '../apis');
 
+/**
+ * Recursively removes all files ending in '-internal.yaml' from a directory and its subdirectories
+ * @param directoryPath - The path to the directory to process
+ */
+function removeInternalOas(directoryPath: string): void {
```
alexvuong, line 28: Nit: Do we want to cut this function short by expecting a directory to have like 3 depth. If it goes beyond that, we error out? I am thinking of a scenario where someone accidentally created lik 10 nested folder structure, and this func will try to run them all. 😬

src/static/helpers/slasHelper.ts
```diff
@@ -226,7 +226,7 @@ export async function authorizeIDP(options: {
     organizationId: slasClient.clientConfig.parameters.organizationId,
     shortCode: slasClient.clientConfig.parameters.shortCode,
   };
-  const queryParams: ShopperLoginQueryParameters = {
+  const queryParams: authorizeCustomerQueryParameters = {
```
vcua-mobify, line 229: Was this an API change? Is this breaking?

dragonza (approved):

PR #210, opened 2025-08-12
Feat: Add API versions to CHANGELOG (@W-18670359@)

joeluong-sfcc:


vcua-mobify (approved):
+1

Small nit - not a blocker since this is probably overcomplicating things:
The command works, but maybe a warning in case the changelog is trying to downgrade a version might be a nice to have?

Screenshot 2025-08-12 at 2 28 25 PM

I had an older version of the APIs when I ran `npm run generateVersionTable` and saw that this command added new lines with the older versions.

For now, we should just make it a part of the review process that when the command is run, we make sure that we always have the latest versions of the API.

PR #214, opened 2025-09-02
Fix: require is not defined node-fetch (@W-19399622@)

joeluong-sfcc:
With the latest major version release, we dropped official support for older node versions and only support node versions 20 & 22. With these versions, node exposes a native fetch function which means we can replace our `node-fetch` polyfill.  This helps resolve an issue with one of our other projects that uses ESM and runs into the following error due to us using `require('node-fetch')` in `environment.ts`:
```
ReferenceError: require is not defined
```

## Changes
1. The `doFetch` function in `src/static/helpers/fetchHelper.ts` that is used by all endpoints will be calling the node/browser fetch instead of the one defined in `src/static/helpers/environment.ts`
    - We unfortunately can't remove the `fetch` function and other unused variables in `environment.ts` as we've already exposed them and removing them would technically be a breaking change. I've added a note to remove them for the next major version bump.
3. Due to the change above, our current testing environments don't expose a fetch function and tests run into an error where no fetch function is defined, so I've updated the test environments to run a setup file to add a fetch polyfill
    - I took a quick stab at trying to just update all of our test dependencies to be on more recent versions to see if newer packages expose a fetch function that the test suites could use, but I think the work is more than expected and so I decided to go with the polyfill approach for now and we can create a follow up ticket to update the testing dependencies

## How to test

You can test this by making API calls both on the browser and the server and ensure things work as expected.

1. In your local version of isomorphic SDK: `git checkout ju/fix-node-fetch`
2. `yarn install && yarn renderTemplates && yarn build:lib`

### Browser

Reach out to me for details as we'll be testing with the new project

### Server

1. If you don't already have a project set up for testing the SDKs, create a new node project with `npm init -y`
4. Add a local reference to the `commerce-sdk-isomorphic` in your `package.json`. The example below uses an absolute path, you can also use relative pathing or symlink with `npm link`
5. `npm install` in testing project

```
  "dependencies": {
    "commerce-sdk-isomorphic": "/Users/joel.uong/Desktop/workspace/commerce-sdk-isomorphic",
  }
```

3. Add an `index.js` file and add the following code, replacing the client config variables:

```javascript
const { helpers, ShopperLogin, ShopperSearch } = pkg;

// demo client credentials, if you have access to your own please replace them below.
const CLIENT_ID = "<INSERT_CLIENT_ID>";
const ORG_ID = "<INSERT_ORG_ID>";
const SHORT_CODE = "<INSERT_SHORT_CODE>";
const SITE_ID = "<INSERT_SITE_ID>";

// must be registered in SLAS. On server, redirectURI is never called
const redirectURI = "http://localhost:3000/callback";

const clientConfig = {
  parameters: {
    clientId: CLIENT_ID,
    organizationId: ORG_ID,
    shortCode: SHORT_CODE,
    siteId: SITE_ID,
  },
};

const slasClient = new ShopperLogin(clientConfig);

const guestTokenResponse = await helpers
  .loginGuestUser({
    slasClient, 
    parameters: { redirectURI }
  })
  .then((guestTokenResponse) => {
    console.log("Guest Token Response: ", guestTokenResponse);
    return guestTokenResponse;
  })
  .catch((error) =>
    console.log("Error fetching token for guest login: ", error)
  );

console.log("guestTokenResponse", guestTokenResponse);

const shopperSearch = new ShopperSearch({
  ...clientConfig,
  headers: {authorization: `Bearer ${guestTokenResponse.access_token}`}
});


const searchResults = await shopperSearch.productSearch({
  parameters: {
      q: 'mens shirt',
  }
})

console.log("searchResults: ", searchResults)
```

4. Run `node index.js &> out.txt` and expect that `out.txt` has the expected output

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,13 +51,15 @@ In the next major version release, the SDK will encode special characters (UTF-8
 
 ### Requirements
 
-- Node `^12.x`, `^14.x`, `^16.x`, `^18.x`
+- Node `^20.x` or `^22.x`
 - The SDK requires B2C Commerce API (SCAPI) to be configured. For more info see [Getting started with SCAPI](https://developer.salesforce.com/docs/commerce/commerce-api/guide/get-started.html).
 
 ### Installation
 
 ```bash
-npm install commerce-sdk-isomorphic
+# This package uses yarn, if you don't have yarn:
+# npm install -g yarn
+yarn install commerce-sdk-isomorphic
 ```
 
 ### Usage
--- a/src/static/helpers/environment.ts
+++ b/src/static/helpers/environment.ts
@@ -20,24 +20,17 @@ export const isNode =
   typeof process.versions === 'object' &&
   typeof process.versions.node === 'string';
 
-export const globalObject = isBrowser ? window : globalThis;
+export const globalObject = isBrowser ? window : globalThis || global;
 
 export const hasFetchAvailable = typeof globalObject.fetch === 'function';
 
 // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
 export const fetch: FetchFunction = (() => {
-  if (isNode) {
-    // .default is added because the newer versions of babel doesn't get the default export automatically for require().
-    // eslint-disable-next-line global-require, @typescript-eslint/no-unsafe-return, @typescript-eslint/no-var-requires, @typescript-eslint/no-unsafe-member-access
-    return require('node-fetch').default;
-  }
-
-  // difficult to test in node environment
+  // Difficult to test in node environment
   /* istanbul ignore next */
-  if (!hasFetchAvailable)
-    throw new Error(
-      'Bad environment: it is not a node environment but fetch is not defined'
-    );
+  if (!hasFetchAvailable) {
+    throw new Error('Please use Node.js 18+ for native fetch support.');
+  }
 
   return globalObject.fetch;
 })();
--- a/src/static/helpers/environment.ts
+++ b/src/static/helpers/environment.ts
@@ -20,7 +20,7 @@ export const isNode =
   typeof process.versions === 'object' &&
   typeof process.versions.node === 'string';
 
-export const globalObject = isBrowser ? window : globalThis || global;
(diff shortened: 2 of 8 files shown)


src/static/helpers/environment.ts
```diff
@@ -24,20 +24,14 @@ export const globalObject = isBrowser ? window : globalThis;
...
-
-  // difficult to test in node environment
+  // Difficult to test in node environment
   /* istanbul ignore next */
-  if (!hasFetchAvailable)
-    throw new Error(
-      'Bad environment: it is not a node environment but fetch is not defined'
-    );
+  if (!hasFetchAvailable) {
+    throw new Error('Please use Node.js 18+ for native fetch support.');
```
shethj, line 33: Should we still be asking folks to use 18+ instead of 20 given this is for avoiding breaking changes but we technically don't support Node 18 ?

alexvuong (approved):
Confirmed this PR fixed the issue on Oddysey. We no longer have to transform node-fetch in Oddysey with this fix. Thanks for working on it

README.md
```diff
@@ -51,13 +51,15 @@ In the next major version release, the SDK will encode special characters (UTF-8
...
+- Node `^20.x` or `^22.x`
 - The SDK requires B2C Commerce API (SCAPI) to be configured. For more info see [Getting started with SCAPI](https://developer.salesforce.com/docs/commerce/commerce-api/guide/get-started.html).
 
 ### Installation
 
 ```bash
-npm install commerce-sdk-isomorphic
+# This package uses yarn, if you don't have yarn:
+# npm install -g yarn
+yarn install commerce-sdk-isomorphic
```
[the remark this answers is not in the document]
joeluong-sfcc: We've always been on yarn, it just was never properly documented

src/static/helpers/environment.ts
joeluong-sfcc: Node 18 is the first node version that exposes a native fetch function to node. I used node 18 instead of node 20/22 so that we don't have to manually update this error message every time we drop support for an older node version

If you feel strongly about it I can update it however

PR #218, opened 2025-09-15
feat(design): add messaging api @W-19261824

steelsojka:
Moves the design layer messaging api from the page designer repo to the commerce sdk.


src/design/messaging-api/api-types.ts
```diff
@@ -0,0 +1,663 @@
...
+   * @stability development
+   *
+   * @example
+   * ```typescript
+   * api.unselectComponent({
+   *   componentId: 'component-123'
+   * });
+   * ```
+   */
+  unselectComponent(event: Domain.ComponentUnselectedEvent): void;
```
rfding, line 320: Is there a reason we went with unselect instead of deselect? Just curious since I don't think I've really heard unselect as a term

steelsojka: I didn't even think about. Deselect makes more sense and now is the time to change it.

steelsojka: renamed

PR #223, opened 2025-09-25
feat(design): add more interactions and cleanup @W-19284040

steelsojka:


yarn.lock
mjuraschik, line 1: Why is there a package and a yarn lock file?

src/design/react/components/ComponentDecorator.tsx
```diff
@@ -18,75 +24,67 @@ import {ComponentDecoratorProps} from './component.types';
...
-      [isDesignMode, setSelectedComponent, componentId]
+      [setSelectedComponent, componentId]
     );
 
-    const isSelected = selectedComponentId === componentId;
-    const isHovered = hoveredComponentId === componentId;
-    const showFrame = isSelected || isHovered;
-    const classNames = [];
+    const handleDelete = useCallback(() => {
+      // TODO: Determine if we can get this information in page designer instead of having to pass it.
```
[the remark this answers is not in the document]
steelsojka: I added those ids to the payload thinking it would be easier since the component would know it and it would make it easier on the page designer side, however right now only a couple properties are passed in. Name, isFragment, and the component id. We would have to pass the region id and the parent component id to the design component as well in order to provide that information.

yarn.lock
steelsojka: I have removed it.

PR #228, opened 2025-10-14
feat: route mapping page designer

mjuraschik:


src/design/messaging-api/client.ts
```diff
@@ -99,6 +100,15 @@ export function createClientApi({
             .then(() => {
               isReady = true;
               messenger.emit('ClientReady', {clientId: id});
+
+              // Emit PageTypeMapReady event after ClientReady if pageTypeMap is provided
+              if (pageTypeMap && Object.keys(pageTypeMap).length > 0) {
+                messenger.emit('PageTypeMapReady', {
```
mjuraschik: put that here instead
```messenger.emit(
          'ClientInitialized',
          {clientId: id, forwardedKeys},
          {requireRemoteId: false}
        );
```

PR #238, opened 2025-10-28
feat: Shopper Consents OAS v1.1.3

mitesh-patel-crm:
Upgrade from v1.1.0 to v1.1.3 for Shopper Consents OAS spec.

### DIFF
--- a/.depcheckignore
+++ b/.depcheckignore
@@ -6,3 +6,8 @@
 /apis
 /docs
 /node_modules
+
+# Ignore as it dynamically imports the commerce-sdk-isomorphic package when utility is used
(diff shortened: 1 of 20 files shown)


.github/workflows/publish-preview.yml
```diff
@@ -0,0 +1,32 @@
+name: Nightly Preview Publish
```
joeluong-sfcc: This isn't required for the preview release, but it's fine to merge this in

joeluong-sfcc (approved):
confirmed that SDK generation works and that new endpoint `updateSubscriptions` is generated

PR #240, opened 2025-10-28
Design layer implementation

steelsojka:
This merges the feature branch for the design layer integration into the preview branch.

### DIFF
--- a/src/design/react/components/DesignRegion.tsx
+++ b/src/design/react/components/DesignRegion.tsx
@@ -4,7 +4,7 @@
  * SPDX-License-Identifier: BSD-3-Clause
  * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/BSD-3-Clause
  */
-import React from 'react';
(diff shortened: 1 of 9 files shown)


mjuraschik (approved):
skimmed over the files, since we reviewed all the PRs separately. nothing jumps out besides the missing codeowners entry

kzheng-sfdc:
Will there be a follow up PR to integrate the design code from the storefront template, or is that already done? 
If a partner created their own template from scratch, how much effort do they need in order to make their template compatible with PD by including the design library?

mjuraschik:
@kzheng-sfdc there will be a follow up to move the component registry code here, which was the initial plan but due to the struggle of integrating the changes from here without releasing a version, we didn't do that right away.
Partners creating their own template are going to be able to use the component (and region) registry in their storefront implementation. What they would need to implement on top of this, is using the registries in their components and pages. We might want to think about releasing the region and component react components as part of this library too to make that easier.

PR #278, opened 2026-03-12
@W-21076912: Feat: Support enableHttpOnlySessionCookies in SLAS helper functions

unandyala:
When enableHttpOnlySessionCookies is true, the PWA-KIT middleware strips tokens from the SLAS response body and sets them as httpOnly cookies. During SSR, the SDK needs to extract these tokens from Set-Cookie headers and merge them back into the TokenResponse.

Changes:
- Add extractTokensFromResponse (private) to parse Set-Cookie headers and map cookie names (cc-at_, cc-nx_, cc-nx-g_, idp_access_token_) back to TokenResponse fields
- Add optional enableHttpOnlySessionCookies flag (defaults to false) to: loginGuestUser, loginGuestUserPrivate, loginIDPUser, loginRegisteredUserB2C, refreshAccessToken, getPasswordLessAccessToken
- When flag is true: call getAccessToken with rawResponse=true and extract tokens from Set-Cookie headers (server-side only)
- When flag is false: existing behavior unchanged (no rawResponse)
- isBrowser check ensures tokens are never exposed on the client

Testing locally with PWA-KIT

  1. Build the SDK:
```
  cd ~/pwa/commerce-sdk-isomorphic
  yarn run build:lib && yarn pack
```
  2. Checkout feature branch [feature/httponly-session-cookies](https://github.com/SalesforceCommerceCloud/pwa-kit/tree/feature/httponly-session-cookies) and enable slas private client.
  3. In pwa-kit/packages/commerce-sdk-react/package.json:
 ` "commerce-sdk-isomorphic": "file:<path-to>/commerce-sdk-isomorphic/package.tgz"`
  4. Install from root: cd ~/pwa/V3/pwa-kit && npm install
  5. Enable private client and httponly session cookies and verify SSR requests are successful.

https://github.com/user-attachments/assets/a20a3b92-85c9-4e41-9d25-c79b62025b5f

### DIFF
--- a/src/static/helpers/slasHelper.ts
+++ b/src/static/helpers/slasHelper.ts
@@ -23,6 +23,89 @@ import {
   CustomRequestBody,
 } from './types';
 
+/**
+ * Parses a Set-Cookie header value and returns the cookie name and value.
+ */
+const parseSetCookieNameValue = (
+  cookie: string
+): {name: string; value: string} | null => {
+  const [nameValue] = cookie.split(';');
+  const eqIndex = nameValue.indexOf('=');
+  if (eqIndex === -1) return null;
+  return {
+    name: nameValue.substring(0, eqIndex).trim(),
+    value: nameValue.substring(eqIndex + 1).trim(),
+  };
+};
+
+/**
+ * Extracts SLAS tokens from Set-Cookie response headers.
+ * Maps httpOnly cookie names back to TokenResponse fields.
+ * @param setCookieHeaders - Array of Set-Cookie header strings
+ * @param siteId - The site ID used as the cookie suffix
+ * @returns Partial TokenResponse with extracted token fields
+ */
+const extractTokensFromSetCookieHeader = (
+  setCookieHeaders: string[],
+  siteId: string
+): Partial<TokenResponse> => {
+  const tokens: Partial<TokenResponse> = {};
+  setCookieHeaders.forEach(cookie => {
+    const parsed = parseSetCookieNameValue(cookie);
+    if (!parsed) return;
+
+    const {name, value} = parsed;
+    if (name === `cc-at_${siteId}`) {
+      tokens.access_token = value;
+    } else if (name === `cc-nx_${siteId}` || name === `cc-nx-g_${siteId}`) {
+      tokens.refresh_token = value;
+    } else if (name === `idp_access_token_${siteId}`) {
+      tokens.idp_access_token = value;
+    }
+  });
+  return tokens;
+};
+
+/**
+ * Parses a SLAS token response. When enableHttpOnlySessionCookies is true
+ * and running on the server, tokens are extracted from Set-Cookie headers
+ * and merged back into the response body (since the middleware strips them).
+ * On the browser, the stripped body is returned as-is (tokens are handled
+ * via httpOnly cookies automatically).
+ * @param response - The raw Response from a SLAS token endpoint
+ * @param siteId - The site ID used as the cookie suffix
+ * @param enableHttpOnlySessionCookies - Whether httpOnly session cookies mode is enabled
+ * @returns A complete TokenResponse
+ */
+const parseTokenResponseWithHttpOnlySessionCookies = async (
+  response: Response,
+  siteId: string,
+  enableHttpOnlySessionCookies = false
+): Promise<TokenResponse> => {
+  const text = await response.text();
+  const body = (text ? JSON.parse(text) : {}) as TokenResponse;
+
+  if (!enableHttpOnlySessionCookies || isBrowser) {
+    return body;
+  }
+
+  // Server-side + httpOnly enabled: restore tokens from Set-Cookie headers
+  // getSetCookie() is the standard API but may not exist in older runtimes
+  const {getSetCookie} = response.headers as unknown as {
+    getSetCookie?: () => string[];
+  };
+  const setCookieHeaders: string[] =
+    getSetCookie?.call(response.headers) ??
+    response.headers.get('set-cookie')?.split(/,(?=\s*\w+=)/) ??
+    [];
+
+  const cookieTokens = extractTokensFromSetCookieHeader(
+    setCookieHeaders,
+    siteId
+  );
+  return {...body, ...cookieTokens};
+};
+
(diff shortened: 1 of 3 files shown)


src/static/helpers/slasHelper.ts
```diff
@@ -312,9 +396,23 @@ export async function loginIDPUser(options: {
       },
       body: tokenBody,
     };
+    if (enableHttpOnlySessionCookies) {
```
joeluong-sfcc: We reuse this logic quite a bit:

```
if (enableHttpOnlySessionCookies) {
  const response = await slasClient.getAccessToken(opts, true);
  return parseTokenResponseWithHttpOnlySessionCookies(
    response,
    slasClient.clientConfig.parameters.siteId
  );
}
```

Can we pull this into a helper function and re-use it?

src/static/helpers/slasHelper.ts
```diff
@@ -23,6 +23,84 @@ import {
...
+  const tokens: Partial<TokenResponse> = {};
+  setCookieHeaders.forEach(cookie => {
+    const parsed = parseSetCookieNameValue(cookie);
+    if (!parsed) return;
+
+    const {name, value} = parsed;
+    if (name === `cc-at_${siteId}`) {
+      tokens.access_token = value;
+    } else if (name === `cc-nx_${siteId}` || name === `cc-nx-g_${siteId}`) {
+      tokens.refresh_token = value;
```
joeluong-sfcc, line 61: Would there every be a case where both `cc-nx_${siteId}` and `cc-nx-g_${siteId}` cookies be sent in the same response? If so, there will be a collision between the two where `tokens.refresh_token` will be set to the cookie that's processed last

src/static/helpers/slasHelper.ts
```diff
@@ -23,6 +23,84 @@ import {
...
+  if (isBrowser) {
+    return body;
+  }
+
+  // Server-side + httpOnly enabled: restore tokens from Set-Cookie headers
+  // node-fetch's Headers has a raw() method that returns individual Set-Cookie headers as an array
+  const headers = response.headers as Headers & {
+    raw?: () => Record<string, string[]>;
+  };
+  const setCookieHeaders: string[] = headers.raw?.()?.['set-cookie'] ?? [];
```
joeluong-sfcc: `.raw()` is specific to the node-fetch polyfill, but we also allow customers to pass in their own custom fetch implementation, and if we upgrade to use native node fetch in the future, we'd have to update this block as well, so let's futureproof it by checking if `getSetCookie` (native node fetch's way to get the headers as an array) is available:

```suggestion
const setCookieHeaders: string[] =
  headers.getSetCookie?.() ??
  headers.raw?.()?.['set-cookie'] ??
  [];
```

src/static/helpers/slasHelper.ts
```diff
@@ -688,6 +841,24 @@ export async function getPasswordLessAccessToken(options: {
     code_verifier: codeVerifier,
     ...(parameters.dnt && {dnt: parameters.dnt}),
   };
+  if (enableHttpOnlySessionCookies) {
+    const response = await slasClient.getPasswordLessAccessToken(
```
joeluong-sfcc, line 841: We duplicate the request options for both `slasClient.getPasswordLessAccessToken` calls, let's pull this out into a variable to re-use

src/static/helpers/slasHelper.ts
```diff
@@ -23,6 +23,84 @@ import {
...
+ * via httpOnly cookies automatically).
+ * @param response - The raw Response from a SLAS token endpoint
+ * @param siteId - The site ID used as the cookie suffix
+ * @returns A complete TokenResponse
+ */
+const parseTokenResponseWithHttpOnlySessionCookies = async (
+  response: Response,
+  siteId: string
+): Promise<TokenResponse> => {
+  const text = await response.text();
```
joeluong-sfcc, line 86: response errors aren't currently caught with this current implementation

```suggestion
  if (!response.ok) {
    throw new ResponseError(response);
  }
  const text = await response.text();
```

joeluong-sfcc (commented):
I haven't tested this yet locally, but leaving some comments for now, other than the comments implementation LGTM

src/static/helpers/slasHelper.ts
unandyala: cc-nx-g_ is deleted when cc-nx_ cookie is added. So there will be only one cookie at a time

src/static/helpers/slasHelper.ts
unandyala: I had this before but then removed it. I will add it back

joeluong-sfcc (dismissed):
LGTM, thanks for addressing PR feedback

src/static/helpers/slasHelper.ts
```diff
@@ -23,6 +23,104 @@ import {
...
+ * Parses a SLAS token response. When enableHttpOnlySessionCookies is true
+ * and running on the server, tokens are extracted from Set-Cookie headers
+ * and merged back into the response body (since the middleware strips them).
+ * On the browser, the stripped body is returned as-is (tokens are handled
+ * via httpOnly cookies automatically).
+ * @param response - The raw Response from a SLAS token endpoint
+ * @param siteId - The site ID used as the cookie suffix
+ * @returns A complete TokenResponse
+ */
+const parseTokenResponseHttpOnly = async (
```
vcua-mobify, line 79: Based on the comment, this assumes we're running behind the PWA middleware.

What happens if someone isn't using the PWA / is using the isomorphic library directly?

unandyala: @vcua-mobify The changes are applied only when `enableHttpOnlySessionCookies=true` is passed to the helper functions. By default `enableHttpOnlySessionCookies=false` and does not impact existing usage/customers

vcua-mobify: We might want to clarify that `enableHttpOnlySessionCookies=true` is only meant to be used with PWA. If not using PWA, leave it as the default `enableHttpOnlySessionCookies=false`

vcua-mobify (approved):

joeluong-sfcc (approved):

PR #282, opened 2026-04-23
[feat] W@22204269 Release v5.2.0 - Add maintenance mode detection

priandsf:
Add automatic maintenance mode detection via throwOnMaintenanceHeader option.

When enabled, the SDK detects maintenance windows by checking the sfdc_maintenance response header and throws MaintenanceError (503) when the value is 'system' or 'site'.

Key features:
- Opt-in via client configuration (backward compatible)
- New MaintenanceError class with detailed error information
- Works with all API endpoints
- Takes precedence over other error handling
- Comprehensive test coverage (100%)

Changes:
- Add MaintenanceError class and tests
- Add throwOnMaintenanceHeader to ClientConfig
- Add header check in doFetch()
- Export MaintenanceError from helpers
- Update README with usage documentation
- Update CHANGELOG for v5.2.0
- Bump version to 5.2.0


src/static/clientConfig.ts
```diff
@@ -89,6 +92,7 @@ export default class ClientConfig<Params extends BaseUriParameters>
       this.proxy = config.proxy;
     }
     this.throwOnBadResponse = !!config.throwOnBadResponse;
+    this.throwOnMaintenanceHeader = !!config.throwOnMaintenanceHeader;
```
unandyala, line 95: Why not just use throwOnBadResponse flag instead of using new flag. when throwOnBadResponse=true, the current code already throws error. We can throw MaintenanceError in this case

unandyala (dismissed):
overall the changes look good to me. I just left one comment

src/static/clientConfig.ts
priandsf: This is actually on purpose to maintain backward compatibility. One may just enable this option while not having exceptions for the other errors.

kevinxh (approved):

ddiazccrz (approved):

PR #286, opened 2026-06-08
Feat: Support named imports via exports map (@W-22860508@)

joeluong-sfcc:
## Summary
- `commerce-sdk-isomorphic` had no `exports` field, so bundlers and tooling that resolve `exports` (and TypeScript `node16`/`nodenext`) fell back to the UMD `main` and couldn't see named exports — `import { ShopperLogin } from 'commerce-sdk-isomorphic'` failed, forcing consumers to default-import and destructure.
- This adds a root `exports` map whose `import`/`require`/`types` conditions point at the **existing** ESM/CJS bundles and declarations. No bundle output changes, no source changes.
- A prior attempt added named imports bolted onto per-API subpath bundles that ~doubled package size and was reverted. This restores named imports **only** — no subpath entries, no bundle-size change (`.tgz` stays 1.09 MB < the 1.2 MB guard).

```js
// New — named imports now resolve
import { ShopperLogin, helpers } from 'commerce-sdk-isomorphic';

// Old — default import + destructure still works (unchanged)
import sdk from 'commerce-sdk-isomorphic';
const { ShopperLogin, helpers } = sdk;
```

## Test plan
- [x] `node --test src/test/packageExports.mjs` — new resolution guard, proven red when the `exports` field is removed and green with it. Runs in the `test` chain after `build:lib`.
- [x] `yarn check:types`, `yarn lint` (0 errors), `yarn test:unit` (221/221), `yarn check:size` (1.09 MB < 1.2 MB) all green.
- [x] Cross-repo smoke: built locally, linked into `pwa-kit`'s `commerce-sdk-react` (which uses named imports throughout), booted the retail storefront, and confirmed the PLP renders products via `ShopperSearch` with no SDK resolution errors.

## Verification
- Fixes named imports for bundler / `exports`-honoring consumers (the common case). Does NOT enable raw `node` native ESM (no bundler): the ESM bundle ships as `.js` (CommonJS without `"type":"module"`), and `environment.ts` resolves `node-fetch` via an eager `require` that throws under native ESM. Closing that needs a `.mjs` build + native-`fetch` migration — out of scope (the bundling surface that triggered the prior revert).

## Out of scope
- Per-API subpath exports (`commerce-sdk-isomorphic/shopperLogin`) — the reverted feature; reintroduces the size regression.
- `.mjs` ESM output + dropping `node-fetch` for native `fetch` — see Verification.

## Follow-ups
- `test:exports` requires `lib/` built first; CI and the pre-push hook build before testing, but a bare local `yarn test` without a prior `build:lib` will fail the exports test.
- `engines.node` still declares `>=12` while `node --test` / `import.meta.resolve` require Node 18+ — pre-existing.

## Ticket
W-22860508

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,6 +11,15 @@
     "url": "git+https://github.com/SalesforceCommerceCloud/commerce-sdk-isomorphic.git"
   },
   "license": "BSD-3-Clause",
+  "exports": {
+    ".": {
+      "types": "./lib/index.cjs.d.ts",
+      "import": "./lib/index.esm.js",
+      "require": "./lib/index.cjs.js"
+    },
+    "./lib/*": "./lib/*",
+    "./package.json": "./package.json"
+  },
   "main": "lib/index.cjs.js",
   "module": "lib/index.esm.js",
   "style": "lib/default.css",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -11,6 +11,15 @@
     "url": "git+https://github.com/SalesforceCommerceCloud/commerce-sdk-isomorphic.git"
   },
   "license": "BSD-3-Clause",
+  "exports": {
```
unandyala, line 14: Does this break existing customers if they have explicit deep imports?

joeluong-sfcc: Good question! No, deep imports into the published package still works. The `exports` map lists `./lib/*`, which is a passthrough for everything under `lib/`, So `import 'commerce-sdk-isomorphic/lib/…'` resolves exactly as before. I verified `require('commerce-sdk-isomorphic/lib/index.cjs.js')` still resolves on this branch.

unandyala: Thanks for  verifying!

unandyala (approved):

PR #289, opened 2026-06-18
Feat: Auto tag and release on release-PR merge (@W-23051165@)

joeluong-sfcc:
## Summary
- Closes the manual step at the tail of the SDK release pipeline: `publish.yml` already chains off `release: published` to run `npm publish`, but the GitHub release that triggers it is still created by hand.
- Fires only on a release-PR merge from a `release/v<X.Y.Z>` branch that passes a three-source version cross-check (head ref vs `package.json` vs `CHANGELOG.md`), so a misfire on a non-release merge can't happen without all three sources also being wrong.
- npm publish stays manual per AC; this PR only closes the release-creation loop.

## Acceptance criteria
- [x] Triggers on `pull_request: closed` to `main`, gated on `merged == true` and `head.ref` starting with `release/v` (the orchestrator-owned release-branch convention). Non-release merges no-op via the job-level `if:`.
- [x] Tag is created at the merge commit (`merge_commit_sha`), format `vX.Y.Z` matching prior tags on the repo.
- [x] GitHub release created and linked to the tag.
- [x] Release notes pulled from `CHANGELOG.md` — only the section for the version being released, awk-bounded between this version's `## v…` header and the next.
- [x] Release name matches the version (e.g. `v3.4.0`).
- [x] npm publish explicitly out of scope.
- [x] Existing release Action surveyed (`publish.yml` runs `npm publish` on `release: published`); the new workflow chains in front of it rather than replacing.

## How the gate composes
The job-level `if:` short-circuits unless the PR's source branch starts with `release/v`. The version step's bash regex anchors on the whole `head.ref` (`^release/v(X\.Y\.Z)$`) — strict enough to reject `release/v5.4.0-rc.1`, `release/v5.4`, `release/v5.4.0/foo`, and same-day re-cut suffixes like `release/v5.4.0-2`. The captured semver becomes one of three sources the cross-check requires to agree.

## Producer–consumer pairing
This workflow's gate consumes a branch convention the orchestrator side produces — the local `release-sdk` skill (Step 3) cuts `release/v$NEW_VERSION`. The skill update ships in lockstep with this PR; if either side regresses (e.g. a hand-cut `release/2026...` branch), the workflow no-ops loudly rather than silently shipping a wrong tag.

## Test plan
- [x] `node --test src/test/releaseOnMergeWorkflow.mjs` — 14 cases passing locally. The harness reads the workflow YAML, extracts each `run: |` block, and pipes it through `bash` with the same env vars the runner sets, so a regex weakened in the YAML reflows the test results here too. Wired into `yarn test` via a new `test:workflow` script.
- [x] Sensitivity proof — temporarily weakened the head-ref regex to `^release/v(.+)$` in the YAML, re-ran the suite: 4 of 14 tests went red (the `-rc.1`, trailing-path, truncated semver, and same-day re-cut rejections). Restored, all green again.
- [x] Live runner verification (Phase 5, branch-only dispatch harness — added in `e6628ab`, removed in `f181a18`, pre-gate-swap):
  - Push trigger with empty `PR_TITLE` → workflow `failure` at the regex step (correct).
  - Dispatch with bad title format → `failure` at the regex step (correct).
  - Dispatch with version mismatch → `failure` at the cross-check step with all three values printed.
  - Dispatch with happy path → `success`. Notes extracted exactly as the v5.4.0 CHANGELOG section, heredoc delimiter survived the step-output → next-step-env round trip.
- [x] `actionlint` clean against the updated `release-on-merge.yml`.

## Verification
**What the local + dispatch checks cover:**
- Head-ref regex on real-runner bash (10 positive/negative cases).
- Three-source version cross-check (head.ref ↔ `package.json` ↔ `CHANGELOG.md`), including the all-three-values error printout.
- awk extraction against multi-version, padded, and missing-version CHANGELOG fixtures.
- Heredoc roundtrip through `\$GITHUB_OUTPUT` with a unique delimiter.
- Step-output → next-step-env propagation on a real GitHub Actions runner (proven on the original PR's pre-gate-swap dispatch runs; the gate-swap diff renames `PR_TITLE` to `HEAD_REF` but the wiring is unchanged).

**What it does NOT cover:**
- The actual `git push origin <tag>` against the live remote.
- The actual `gh release create` against the live repo.

**Plan to close the remaining gaps:**
- First real exercise is the next release-PR merge. Risk surface is bounded by the strict head-ref regex + three-source cross-check + skip-if-exists on tag and release. A misfire produces a loud workflow `failure`, never a wrong tag.
- Reviewer ask: please confirm the repo's tag protection rules (if any) don't bar `github-actions[bot]` from pushing tags matching `v*`. If they do, the workflow will go red on the first real release and we'll need a deploy-time configuration fix.

## TODO before merge
- [x] Delete the dispatch dry-run workflow (`release-on-merge-dryrun.yml`) — done in `f181a18`. The cumulative diff is clean; the two intermediate commits are noise on the branch's history but not in the PR's net change.

## Out of scope
- npm publish automation — separate ticket per AC. `publish.yml` continues to run on `release: published`.
- Backfilling tags / releases for already-shipped versions.
- Resetting the `preview` branch — the `release-sdk` orchestrator already opens that as a separate PR.
- Hardening of `publish.yml` itself (e.g. `--ignore-scripts` on `yarn install`, tag-format gating on the `release: published` trigger) — surfaced by the security review of this branch but pre-existing on `main`; tracked separately.


.github/workflows/release-on-merge.yml
```diff
@@ -0,0 +1,108 @@
...
+  tag-and-release:
+    if: github.event.pull_request.merged == true && startsWith(github.event.pull_request.head.ref, 'release/v')
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.pull_request.merge_commit_sha }}
+          fetch-depth: 0
```
vmarta: Minor: for what this PR does, we don't need to fetch the full git history.

```suggestion
```

```
Why this PR doesn't need it

  The workflow does three git things:

  1. Reads package.json and CHANGELOG.md from the working tree (no history needed).
  2. git tag -a "$TAG" -m "$TAG" "$SHA" — creates a tag pointing at one specific commit (merge_commit_sha). It only needs that one commit to
  exist locally, which fetch-depth: 1 already guarantees because that's the commit being checked out.
  3. git ls-remote --tags origin — talks to the remote, doesn't read local history.
  4. git push origin "$TAG" — pushes the tag, doesn't need local history.

  None of those need ancestor commits. So fetch-depth: 1 (or just dropping the parameter — the default is shallow) would work and shave time
  off every release.
```

vmarta (dismissed):
👍 I see now which manual steps of the release process we're automating here in the PR:

```
  What was manual before

  To cut a real release, a human had to do all of this by hand:

  1. Decide the new version (say 5.4.0).
  2. Cut a branch named release/v5.4.0.
  3. Edit package.json to bump version to 5.4.0.
  4. Edit CHANGELOG.md to add a ## v5.4.0 section with what changed.
  5. Open a PR from release/v5.4.0 → main.
  6. Get it reviewed and merge it.
  7. ← Manual gap starts here.
  8. Go to GitHub's web UI → Releases → "Draft a new release."
  9. Type v5.4.0 as the tag name, point it at the merge commit.
  10. Copy-paste the CHANGELOG section into the release body.
  11. Click "Publish release."
  12. ← Manual gap ends. publish.yml takes over and runs npm publish.

  Steps 8–11 are the part that was tedious, easy to mess up (wrong tag, typo in notes, pointing at the wrong commit), and entirely doable by
  a script. They were the only remaining manual step in the whole chain.

  What this PR automates

  This PR adds a fourth workflow, release-on-merge.yml, that watches for a release-PR getting merged and does steps 8–11 for you.

  Concretely, when a PR merges into main:

  1. It checks: did this PR come from a release/v… branch? If not (it's just some normal feature PR), the workflow exits silently. This is
  the gate.
  2. It cross-checks three sources agree on the version:
    - The branch name says release/v5.4.0 → version is 5.4.0.
    - package.json says "version": "5.4.0" → version is 5.4.0.
    - CHANGELOG.md's top section says ## v5.4.0 → version is 5.4.0.

  If any two disagree, it fails loudly and stops. This is a "three witnesses" sanity check — a misfire would require a human to have got the
  version wrong in three independent places.
  3. It pulls the release notes out of CHANGELOG.md — just the section under ## v5.4.0, stopping when it hits the next ## v… heading.
  4. It creates the git tag v5.4.0 pointing at the merge commit and pushes it.
  5. It creates the GitHub Release with that tag and those notes.
  6. publish.yml then fires automatically (it's already wired to release: published) and runs npm publish.

  So the whole chain becomes: human merges the release PR → everything else happens.

  What's still manual

  The PR is intentionally narrow:
  - A human still cuts the release/v… branch, bumps package.json, and writes the CHANGELOG entry. (There's a separate release-sdk "skill" —
  basically a Claude Code-driven script — that does this part interactively.)
  - A human still reviews and merges the release PR. The merge is the green-light moment; nothing publishes without it.
  - npm publishing itself stays untouched — publish.yml already does it.
```

.github/workflows/release-on-merge.yml
joeluong-sfcc: good catch, the change has been applied

vmarta (approved):

PR #293, opened 2026-07-09
Fix: Auto-publish on release-PR merge  (@W-23235820@)

joeluong-sfcc:
## Summary
- Auto-publish to npm when a `release/v*` branch is merged to `main`, and make sure it works alongside the auto-created GitHub release from `release-on-merge.yml`.
- Both workflows fire off the same `pull_request: closed` event on a merged `release/v*` branch, authenticate with `GITHUB_TOKEN`, and run in parallel — neither waits on the other. Prior design chained `publish.yml` off `release: published`, which GitHub drops when the release is authored by `GITHUB_TOKEN` (bit v5.4.0 — tagged and released fine, never published to npm).
- `workflow_dispatch` stays on `publish.yml` as a manual fallback that takes an existing tag and re-publishes that ref.

## Test plan
- [x] `yarn test:workflow` passes — locking test asserts the new trigger surface (regex-on-YAML): `pull_request: closed` trigger, `merged==true && startsWith(head.ref, 'release/v')` if-condition, `merge_commit_sha` checkout ref, absence of `release: published`, required `workflow_dispatch` tag input, and that `release-on-merge` never reintroduces a PAT for release creation.
- [x] Sensitivity-checked each of the six assertions by temporarily reverting the surface it locks and confirming the test goes red.
- [x] `actionlint .github/workflows/publish.yml .github/workflows/release-on-merge.yml` — clean (the two `actions/checkout@v3` / `setup-node@v3` warnings are pre-existing, unchanged by this diff).

### What the local checks cover
- The YAML surface: trigger, if-condition, checkout ref, workflow_dispatch input, and the `GITHUB_TOKEN` / no-PAT invariant on `release-on-merge`. Locked in CI.
- The REST-payload contract each step reads from: `gh api repos/SalesforceCommerceCloud/commerce-sdk-isomorphic/pulls/291 --jq '{merged, merge_commit_sha, head}'` on the already-merged v5.4.0 PR confirms `merge_commit_sha` is populated and matches the commit `release-on-merge` tagged (annotated tag deref via `git/tags/<obj-sha>`). This is the same commit `publish.yml` would check out on the merge path, so `package.json` version at checkout matches the tag.

### What it does NOT cover
- The workflow-runtime plumbing: does GitHub actually fire `publish.yml` on `pull_request: closed` given this `if:` gate, does the runner resolve `merge_commit_sha` to a checkoutable commit, do `release-on-merge` and `publish.yml` both fire on the same merge event. Not observable from a branch-only dispatch — we can't fake a merged PR without merging one.

### Plan to close the remaining gaps
- First real release (v5.5.0). Expected: both workflows fire in parallel from the same `pull_request: closed` event; `release-on-merge` tags + creates the release with `GITHUB_TOKEN`; `publish.yml` checks out the merge commit and \`npm publish\` succeeds. If `publish.yml` doesn't fire, the `workflow_dispatch` fallback unblocks the release: run it with `tag: v5.5.0` and it checks out the tag `release-on-merge` created.

## Follow-ups
- No `concurrency:` group on `publish.yml`. If a manual `workflow_dispatch` overlaps a merge-triggered run, they'd race on `npm publish`. Not corrupting (npm 409s on duplicate versions), recoverable via re-run. Not worth blocking the pivot; can be added later.

## Out of scope
- Migrating to a GitHub App-based token (`peter-evans/create-github-app-token`). Mooted by the pivot — no PAT means no App-replacement need.
- Trusted Publishing (OIDC) on npm — tracked separately as W-22028201.
- Retroactively fixing the v5.4.0 release chain — already unblocked manually.

## Ticket
W-23235820


.github/workflows/publish.yml
```diff
@@ -1,20 +1,36 @@
...
   publish-to-npm:
+    if: |
+      github.event_name == 'workflow_dispatch' ||
+      (github.event.pull_request.merged == true &&
+       startsWith(github.event.pull_request.head.ref, 'release/v'))
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
     steps:
       - uses: actions/checkout@v3
```
vmarta: Nit (non-blocking): the sibling `release-on-merge.yml` uses `actions/checkout@v4`, but here checkout and `setup-node` are still on `@v3`. Same event, same job shape, different action versions — plus `@v3` triggers the deprecation warnings called out in the PR body. Since you are already editing this file, worth bumping both to `@v4` here rather than a follow-up.

vmarta (approved):
Approve. Root-cause fix, correct and cleanly verified.

**Plain-English summary:** Releases were supposed to auto-publish to npm, but they did not. The old design chained two steps: create the GitHub Release, then publish to npm on the "release published" signal. GitHub suppresses that signal when the release is authored by `GITHUB_TOKEN`, so publish never fired (v5.4.0 tagged + released but never hit npm). This PR stops chaining: both the release workflow and the publish workflow now fire in parallel off the same "release branch merged to main" event, each checking out the same merge commit. Adds a `workflow_dispatch` manual fallback and a test that locks the trigger surface so the old broken design cannot silently return.

One non-blocking nit inline. `concurrency:` group and structural (vs regex) YAML test noted as reasonable follow-ups.

.github/workflows/publish.yml
joeluong-sfcc: Done, both bumped to @v4 to match release-on-merge.yml. Locking test updated too.

PR #295, opened 2026-07-28
Chore: Reset preview to match main for v5.5.0 (@W-23614473@)

joeluong-sfcc:
Resets the `preview` branch to match `main` as part of the v5.5.0 SDK release.
The diff below is everything currently on `preview` that is NOT on `main` —
i.e. the work that will be discarded when this PR merges. If any of
those commits represent unmerged dev work, speak up in this PR before
merging.

**Merge with a merge commit — do NOT squash.** The reset relies on `preview`
and `main` sharing history; squashing collapses that shared history and makes
the next preview reset diff noisier. After the merge, `preview` will point at
the same tree as `main`.

Ticket: W-23614473


vcua-mobify (approved):
Looks like this PR is adding the auto trigger from main to the preview branch.

